PR #3, opened 2019-01-20
Use a Cached ContentLoader in DomainContentResolver

bdunogier:
The `CachedContentLoader` stores loader Content items to avoid reloading them from the Repository.

The execution time is [down 17%](https://blackfire.io/profiles/compare/bea2b8de-6565-4e7d-9231-d04dd33c9946/graph) on the following query, with way fewer DB/SPI cache* queries:
```
{
  media {
    images(sortBy: [_dateModified, _desc]) {
      _url
      name
      image {
        uri
      }
      caption { html5 }
    }
  }
}
```

* In regards to SPI cache, on a simple query of two fields from 3 articles, cache calls are down from ~500 to ~200. With a redis based cache, the impact is fairly important.

### TODO
- [ ] Take mutations into account. Clear all in-memory cache ?

### DIFF
--- a/spec/EzSystems/EzPlatformGraphQL/GraphQL/Resolver/DomainContentResolverSpec.php
+++ b/spec/EzSystems/EzPlatformGraphQL/GraphQL/Resolver/DomainContentResolverSpec.php
@@ -3,6 +3,8 @@
 namespace spec\EzSystems\EzPlatformGraphQL\GraphQL\Resolver;
 
 use eZ\Publish\SPI\Persistence\Content\Type\Handler as ContentTypeHandler;
+use EzSystems\EzPlatformGraphQL\GraphQL\DataLoader\ContentLoader;
+use EzSystems\EzPlatformGraphQL\GraphQL\DataLoader\ContentTypeLoader;
 use EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\SearchQueryMapper;
 use EzSystems\EzPlatformGraphQL\GraphQL\Resolver\DomainContentResolver;
 use eZ\Publish\API\Repository\Repository;
@@ -14,9 +16,11 @@ class DomainContentResolverSpec extends ObjectBehavior
     function let(
         Repository $repository,
         TypeResolver $typeResolver,
-        SearchQueryMapper $searchQueryMapper
+        SearchQueryMapper $queryMapper,
+        ContentLoader $contentLoader,
+        ContentTypeLoader $contentTypeLoader
     ) {
-        $this->beConstructedWith($repository, $typeResolver, $searchQueryMapper);
+        $this->beConstructedWith($repository, $typeResolver, $queryMapper, $contentLoader, $contentTypeLoader);
     }
 
     function it_is_initializable()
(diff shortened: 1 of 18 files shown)


GraphQL/DataLoader/CachedContentLoader.php
```diff
@@ -0,0 +1,49 @@
...
+            $this->loadedItems[$item->id] = $item;
+        }
+
+        return $items;
+    }
+
+    public function findSingle(Criterion $filter): Content
+    {
+        $contentId = $filter->value[0];
+        if ($filter instanceof Criterion\ContentId && isset($this->loadedItems[$contentId])) {
```
andrerom: if this is often the case you can also map top load methods so they are cached on API/SPI side alos when yet in local cache here.


aka:
```
if ($filter instanceof Criterion\ContentId) {
    if (isset($this->loadedItems[$contentId])) {
        return $this->loadedItems[$contentId];
    } else {
        // As of EZP-29899 (7.4.0, and upcoming 7.3.5, 6.13.6, 6.7.10, 5.4.14) we don't need to specify current version
        return $this->loadedItems[$contentId] = $this->contentService->load($contentId);
    }
```


You can use the same trick in find(), however maybe only if you detect* presence of bulk loading api's:
- `loadContentInfoList` : Coming as part of https://github.com/ezsystems/ezpublish-kernel/pull/2529
- `loadContentListByContentInfo`: Added as part of https://github.com/ezsystems/ezpublish-kernel/commit/2a3684893a2714a8c5c07ea1f4f5bffa3fb2f94c

_(or we could expose a simplified `loadContentList($ids)` also to skip the round trips)_

_Side: You should btw have the bulk/multi-load api's in mind when doing design here, design of ContentTypeLoader for instance will explicitly disallow you from being able to take advantage of the recent method to load several content types at once on ContentTypeService._

* _Assuming you aim to make this support several LTS kernel versions to assure broad reach._

bdunogier: Thank you for your feedback, @andrerom, it helps. I'll change single content item loading to what you suggested, as it makes no sense to use the SearchService to load a single item. 

I'm actually not entirely set about the load method itself, though: it was either a unique method, with a flexible argument (in that case, a Criterion), or one method per supported argument. Would you by any chance have arguments against / in favour of one or the other ?

> design of ContentTypeLoader for instance

Actually, this one shouldn't have made it, I'll remove it from the PR for the time being. But I'll keep your remarks in mind for the future design of it.

bdunogier: It's about `findSingle()` above: I'm not sure yet if I want `findSingle()` or `loadById()`, `loadByRemoteId()`, `loadByLocationId()`...

andrerom: Oh, no strong opinion there, but of course if consumer is your own code here:
1. load and findSingle has same semantics, it throws if missing. But findSignal adds unneeded overhead if you anyway end up mapping it to load method(s)
2. The bulk/multi load suggestion still applies, if there is a need to load several, allow it to not force un-needed round trips to backend _(do backfire profiling on server setup where redis is on network to see what I mean)_

bdunogier: > The bulk/multi load suggestion still applies, if there is a need to load several, allow it to not force un-needed round trips to backend 

I have read your suggestions several times, and while I see the idea of those methods, I'm not sure in which case they should apply here. I'll keep it in mind for the `ContentTypeLoader`, but in regards to the Content one, it is not clear. Should I change my usage of `SearchService::findContent()` to take advantage of this or not ?

> (do backfire profiling on server setup where redis is on network to see what I mean)
Actually, I have, on Platform.sh.

andrerom: Let's find an occasion to discuss, I might need to better understand what you are trying to do

bdunogier:
Done, @webhdx.

PR #7, opened 2019-01-25
fixes issue-87: add ParentLocationId to the SearchQueryMapper

crevillo:
ParentLocationId is not taken in account in the SearchQueryMapper. Sorry i created the issue in your other repo @bdunogier


bdunogier:
Thank you @crevillo !

PR #8, opened 2019-01-26
Refactor query mapper

crevillo:
Prof of concept here. 
It refactors the SearchQueryMapper to allow the extension of those criterios without the need to remodify that class again and again. 
Still working on it, but i guess pr is the way to know if i'm the right track :).

### DIFF
--- a/src/DependencyInjection/Compiler/SearchQueryMappersPass.php
+++ b/src/DependencyInjection/Compiler/SearchQueryMappersPass.php
@@ -11,12 +11,12 @@ class SearchQueryMappersPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        if (!$container->has(self::ID)) {
+        if (!$container->has(SearchQueryMapper::class)) {
             return;
         }
 
         $definition = $container->findDefinition(SearchQueryMapper::class);
-        $criteriaTaggedServices = $container->findTaggedServiceIds('ezplatform_graphql.query_input_criterion_visitor');
+        $criteriaTaggedServices = $container->findTaggedServiceIds('ezplatform_graphql.query_input_visitor');
 
         $queryInputVisitors = [];
         foreach ($criteriaTaggedServices as $id => $tags) {
--- a/src/GraphQL/InputMapper/Search/SortBy.php
+++ b/src/GraphQL/InputMapper/Search/SortBy.php
@@ -9,13 +9,17 @@ class SortBy implements QueryInputVisitor
      */
     private $sortClauseVisitors;
 
-    public function __construct(array $sortClauseVisitors)
+    public function __construct(array $sortClauseVisitors = [])
     {
         $this->sortClauseVisitors = $sortClauseVisitors;
     }
 
     public function visit(QueryBuilder $queryBuilder, $value): void
     {
-        // TODO: Implement visit() method.
+        foreach ($value as $inputKey => $inputValue) {
+            if (isset($this->sortClauseVisitors[$inputKey])) {
+                $this->sortClauseVisitors[$inputKey]->visit($queryBuilder, $inputValue);
+            }
+        }
     }
 }
--- a/src/Resources/config/query_input_visitors.yml
+++ b/src/Resources/config/query_input_visitors.yml
@@ -20,7 +20,6 @@ services:
         tags:
             - {name: 'ezplatform_graphql.query_input_visitor', inputKey: 'ParentLocationId'}
 
-
     EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\Criteria\Created:
         tags:
             - {name: 'ezplatform_graphql.query_input_visitor', inputKey: 'Created'}
@@ -35,4 +34,46 @@ services:
 
     EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortBy:
         tags:
-            - {name: 'ezplatform_graphql.query_input_visitor', inputKey: 'SortBy'}
+            - {name: 'ezplatform_graphql.query_input_visitor', inputKey: 'sortBy'}
+
+    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\ContentName:
+        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\SortClauseClass
+        arguments:
+            $sortClauseClass: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\ContentName'
+        tags:
+            - {name: 'ezplatform_graphql.query_sortclause_visitor', inputKey: 'Name'}
+
+    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\ContentId:
+        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\SortClauseClass
+        arguments:
+            $sortClauseClass: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\ContentId'
+        tags:
+            - {name: 'ezplatform_graphql.query_sortclause_visitor', inputKey: 'ContentId'}
+
+    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\DateModified:
+        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\SortClauseClass
+        arguments:
+            $sortClauseClass: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\DateModified'
+        tags:
+            - {name: 'ezplatform_graphql.query_sortclause_visitor', inputKey: 'DateModified'}
+
+    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\DatePublished:
+        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\SortClauseClass
+        arguments:
+            $sortClauseClass: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\DatePublished'
+        tags:
+            - {name: 'ezplatform_graphql.query_sortclause_visitor', inputKey: 'DatePublished'}
+
+    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\SectionIdentifier:
+        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\SortClauseClass
+        arguments:
+            $sortClauseClass: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\SectionIdentifier'
+        tags:
+            - {name: 'ezplatform_graphql.query_sortclause_visitor', inputKey: 'SectionIdentifier'}
+
+    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\SectionName:
+        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\SortClauses\SortClauseClass
+        arguments:
+            $sortClauseClass: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\SectionName'
+        tags:
+            - {name: 'ezplatform_graphql.query_sortclause_visitor', inputKey: 'SectionName'}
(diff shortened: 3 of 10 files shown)


GraphQL/InputMapper/SearchQueryMapper.php
```diff
@@ -6,6 +6,13 @@
 
 class SearchQueryMapper
 {
+    private $criteriaMappers;
+
+    public function __construct($criteriaMappers)
```
bdunogier: You could typehint this as `array`.

GraphQL/InputMapper/SearchQueryMapper.php
```diff
@@ -14,65 +21,17 @@ public function mapInputToQuery(array $inputArray)
...
-        if (isset($inputArray['ContentTypeIdentifier'])) {
-            $criteria[] = new Query\Criterion\ContentTypeIdentifier($inputArray['ContentTypeIdentifier']);
-        }
-
-        if (isset($inputArray['Text'])) {
-            foreach ($inputArray['Text'] as $text) {
-                $criteria[] = new Query\Criterion\FullText($text);
+        foreach ($inputArray as $inputField => $inputValue) {
+            if ($criteriaMapper = $this->existsCriteriaMapperForField($inputField)) {
+                $criteria = array_merge($criteria, $this->criteriaMappers[$inputField]->map($inputValue));
```
bdunogier: It feels a bit weird that you're using a method to check if there is a visitor for that field at line 25, and then use the index from the `criteriaMappers` array. You may as well use `isset` to check if there's a visitor.

GraphQL/InputMapper/SearchQueryMapper.php
```diff
@@ -87,62 +46,45 @@ function ($sortClauseClass) {
...
 
-        if (!isset($targetMap[$dateMetadata])) {
-            echo "Not a date metadata\n";
-            return [];
-        }
+    /**
+     * @param $inputValue
+     * @return array
+     */
+    private function mapSortByPart($inputValue) {
```
bdunogier: Remember to typehint return values as well: `private function mapSortByPart($inputValue): array`. Also, The opening curly bracket for methods gets its own line (I gotta setup CS check).

Resources/config/search_query_mappers.yml
```diff
@@ -0,0 +1,24 @@
+services:
+    BD\EzPlatformGraphQLBundle\GraphQL\InputMapper\Search\Criterion\ContentTypeIdentifier:
+        tags:
+            - {name: 'ezplatform_graphql.search_query_mapper', criteria: 'ContentTypeIdentifier'}
```
bdunogier: Wouldn't `criteria` be more like `inputKey` ?

crevillo:
thanks for the comments. they all make sense. yes, i was not totally happy either, but i think i will be more happy once i try to work on your suggestions. 
Talk soon.

bdunogier:
Remember that we take an iterative approach. We can break BC as long as we haven't hit v1.0. The refactoring of this input mapper is necessary in any case, and the general direction is a good one. Keep it coming :)

One thing that may help designing good code, since you're into DDD these days, is to write PhpSpec tests for the input mapper & visitors. They make bad code difficult to test, and push you towards good design. It would be a good opportunity to add a new skill to your portfolio.

crevillo:
Applied some of the comments in the pr and also in slack. so now we have
* the query mapper uses a query builder
* the query builder has an `addCriterion` method. Visitors will add criterions to the querybuilder and also can set SortBy. 
* the query Builder has also a buildQuery method that returns a Content Query. 

Open to more suggestions! :)

src/DependencyInjection/Compiler/SearchQueryMappersPass.php
```diff
@@ -0,0 +1,34 @@
...
+namespace EzSystems\EzPlatformGraphQL\DependencyInjection\Compiler;
+
+use EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\SearchQueryMapper;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class SearchQueryMappersPass implements CompilerPassInterface
+{
+    const ID = 'EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\SearchQueryMapper';
```
bdunogier: I would probably remove the constant, and use `SearchQueryMapper::class` in the `if` below. Even more so since you use it right after ;)

src/DependencyInjection/Compiler/SearchQueryMappersPass.php
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        $definition = $container->findDefinition(SearchQueryMapper::class);
+        $taggedServices = $container->findTaggedServiceIds('ezplatform_graphql.query_input_visitor');
+
+        $queryInputVisitors = [];
+        foreach ($taggedServices as $id => $tags) {
+            foreach ($tags as $tag) {
+                if (isset($tag['inputKey'])) {
+                    $queryInputVisitors[$tag['inputKey']] = new Reference($id);
```
bdunogier: Note: the advantage of this is that a query input visitor can be overridden using the key. Useful.

src/Resources/config/query_input_visitors.yml
```diff
@@ -0,0 +1,32 @@
+services:
+    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\Criterion\ContentTypeIdentifier:
+        tags:
+            - {name: 'ezplatform_graphql.query_input_visitor', inputKey: 'ContentTypeIdentifier'}
```
bdunogier: We could think of basing the definition of the `ContentSearchQuery` on the defined `inputKey`.

What it implies:
- add a `DomainIterator` that iterates over registered criterion input visitors
- add a worker that adds a field to `ContentSearchQuery` for each of them

The only remaining question is where we get the type of each field from.

src/GraphQL/InputMapper/Search/Criterion/ParentLocationId.php
```diff
@@ -0,0 +1,15 @@
...
+
+use eZ\Publish\API\Repository\Values\Content\Query;
+use EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\QueryBuilder;
+use EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\QueryInputVisitor;
+
+class ParentLocationId implements QueryInputVisitor
+{
+    public function visit(QueryBuilder $queryBuilder, $value): void
+    {
+        $queryBuilder->addCriterion(new Query\Criterion\ParentLocationId($value));
```
bdunogier: Should we refactor all visitors that instanciate a Criterion object to a generic one ? It could take a criterion class as a constructor argument. We could remove a few duplicate files with this.

crevillo: but Field and DateMetadata seems to work a bit different than the others, aren't they?

bdunogier: Indeed !

But you could at least remove `ContentTypeId`, `ContentTypeIdentifier`, `ParentLocationId` and `Text`. Furthermore, there are more criteria that work the same way, we could easily add them if we have a class for it.

`BasicCriterionVisitor` ? `Standard` ?

src/Resources/config/query_input_visitors.yml
bdunogier: In this regard, we should keep in mind that we would ideally need an input-object (like `ContentSearchQuery`) that does NOT have Content Type based fields: when filtering on a given content-type, they can't be used (well, they can, but then we will search for content that is of two types, and we don't have that :)

src/GraphQL/InputMapper/Search/Criterion/ParentLocationId.php
crevillo: i don't know if i follow you here. 

do you mean having a Standard criterion visitor tagged with allo those input keys? 
my doubt is, how to instatiate this standard criterion and tell it to use one criterion or another depending on the inputkey ?

bdunogier: An example:

```
class BasicCriterionVisitor implements QueryInputVisitor
{
    public function __construct($criterionClass)
    {
        $this->criterionClass = $criterionClass;
    }

    public function visit(QueryBuilder $queryBuilder, $value): void
    {
        $queryBuilder->addCriterion(new {$this->criterionClass}($value));
```

```
services:
    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\Criterion\ContentTypeId:
        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\Criterion\Basic
        arguments:
            $criterionClass: 'eZ\Publish\API\Repository\Values\Content\Query\Criterion\ContentTypeId'
        tags:
            - {name: 'ezplatform_graphql.query_input_visitor', inputKey: 'ContentTypeId'}    

    EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\Criterion\ContentTypeIdentifier:
        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\Criterion\Basic
        arguments:
            $criterionClass: 'eZ\Publish\API\Repository\Values\Content\Query\Criterion\ContentTypeIdentifier'
        tags:
            - {name: 'ezplatform_graphql.query_input_visitor', inputKey: 'ContentTypeIdentifier'}    

EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\Criterion\ParentLocationId:
        class: EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\Criterion\Basic
        arguments:
            $criterionClass: 'eZ\Publish\API\Repository\Values\Content\Query\Criterion\ParentLocationId'
        tags:
            - {name: 'ezplatform_graphql.query_input_visitor', inputKey: 'ParentLocationId'}    
```

Does it make more sense ?

crevillo: yes, much more clear now. i thought you were trying to avoid all those service definitions, but you're only trying to avoid php classes :)
i'll work on this.

crevillo:
created the Standard Criterion visitor and removed the unneded ones. i would need some help with the spec now.

crevillo:
added spec for the query builder and modified the searchQueryMapper spec too.

src/GraphQL/InputMapper/Search/SortBy.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search;
+
+use EzSystems\EzPlatformGraphQL\GraphQL\InputMapper\Search\QueryBuilder;
+
+class SortBy implements QueryInputVisitor
```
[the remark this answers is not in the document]
crevillo: but currently `SortBy` are at "the same level" as criterion when building queries with the graphql sintax, right? so, probably, sortby should still be a query input visitor and then we could have sort clauses visitors that could add sortclauses to the query builder?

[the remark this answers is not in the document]
crevillo: @bdunogier i think it makes sense what you suggest, but let me ask if i'm in the right direction. currently we have this in the Search.types.yml file. 
```
SortByOptions:
    type: enum
    config:
        values:
            _contentId:
                value: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\ContentId'
                description: "Sort by content id"
            _name:
                value: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\ContentName'
                description: "Sort by content name"
            _dateModified:
                value: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\DateModified'
                description: "Sort by last modification date"
            _datePublished:
                value: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\DatePublished'
                description: "Sort by initial publication date"
            _sectionIdentifier:
                value: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\SectionIdentifier'
                description: "Sort by content section identifier"
            _sectionName:
                value: '\eZ\Publish\API\Repository\Values\Content\Query\SortClause\SectionName'
                description: "Sort by section name"
            _desc:
                value: 'descending'
                description: "Reverse the previous sorting option"
```
Do you mean changes definition there to have it more similar to the ContentSearchQuery one? n (i mean, not map to class directly in the yml)

PR #4, opened 2019-01-23
Content create, update & delete support through mutations

bdunogier:
### Create content

```
mutation CreateBlogPost {
  createBlogPost(
    languageCode: "eng-GB",
    parentLocationId: 2,
    input: {
      title: "The blog post's title",
      author: [
        {name: "John Doe", email: "johndoe@unknown.net"}
      ],
      body: {
        format: "html",
        input: "<h1>title</h1><p>A paragraph</p><p>Another <b>paragraph</b></p>"
      }
    }
  ) {
    _info { id mainLocationId }
    title
    body { html5 }
  }
}
```

### Update content

```
mutation updateBlogPost(
  $id: ID!,
  $input: BlogPostContentInput!
) {
  updateBlogPost(id: $id, input: $input) {
    _info { id currentVersionNo }
    title
  }
}
```

Update works like create, but without any mandatory field.

A versionNo argument is supported:
- if not specified, a draft is created from the current version
- if specified, it depends on the version's status:
  - if it is a draft, it is used for the update.
  - if it is an archived or published, it is used to create a draft

Then the draft gets published immediately.

### Delete content

```
mutation deleteBlogPost
{
  deleteContent(contentId: 256) {
    id
    contentId
  }
}
```

### File upload support

Example with curl (required, graphiql won't let you upload files):

```
curl -v -X POST \
  http://localhost:8000/graphql \
  -H 'Cookie: eZSESSIDWhatever=cookievalue' \
  -F 'operations={"query":"mutation CreateImage($name: String!, $alternativeText: String!, $file: FileUpload!) { createImage( parentLocationId: 51, input: { name: $name, image: {alternativeText: $alternativeText, file: $file} } ) { _info { id mainLocationId } name image { fileName alternativeText uri } } }","variables":{"file": null, "name": "2nd image upload", "alternativeText": "With generated schema"}}' \
  -F 'map={"0":["variables.file"]}' \
  -F "0"=@/Users/bdunogier/Desktop/screenshot.png
```

### Multi-file upload support

```
mutation BunchOfImages($files: [ImageUpload]!) {
  uploadFiles(locationId: 51, files: $files) {
    files {
      ... on ImageContent {
        name
        image {
          uri
        }
      }
    }
    warnings
  }
}
```

```
curl -v -X POST \
  http://localhost:8000/graphql \
  -H 'Cookie: eZSESSID98defd6ee70dfb1dea416cecdf391f58=1ei7ecn4vu4afdk9hk0k67kme2' \
[...]

PR #9, opened 2019-01-30
Allow to customize the root schema

bdunogier:
Makes the GraphQL container extension look for predefined files path in `app/config/graphql` that will be loaded instead of the Platform ones. It allows the app maintainer to add new items to the schema and handle custom business objects:

- `app/config/graphql/Query.types.yml` can define a `Query` type. It must inherit the `Domain` type.
- `app/config/graphql/Mutation.types.yml` can define a `Mutation` type. Once mutations are merged (#4), it will have to inherit the `DomainContentMutation` type.

See `doc/custom_app_schema.md` for examples.


bdunogier:
@vfaisse after discussing this extension point internally, we believe that being able to add your own schema using the `overblog_graphql` configuration should be sufficient. Could you confirm that it is ? The only difference with that PR is that you need to use a different endpoint (e.g. `/graphql/custom`), and not integrate it into the ezplatform schema. Please let us know.

bdunogier:
@webhdx CS fixed on the newly added files. The other files only have minor issues (phpdoc header, blank lines) that do not come from this particular PR. Maybe we can take care of them in master once this PR is merged (tomorrow if possible).

vfaisse:
Hi @bdunogier, I confirm that I have successed to add my custom schema with overblog_graphql and a new endpoint as explained in the documenation.
Thank you Bertrand, and still good job for this bundle.

bdunogier:
Thank you for the feedback, @vfaisse ! We have merged this PR nonetheless, since it contained other improvements. We may revert the commit that adds this specific extension point, but not immediately :-)

PR #11, opened 2019-03-11
EZP-30283: Implement fetching Object States in GraphQL

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-30283

# Description
It adds new fields to `_repository`:
- `objectStateGroup` (args: `id`) - fetches single Object State Group,
- `objectStateGroups` - fetches all Object State Groups,
- `objectState` (args: `id`) - fetches single Object State,
- `objectStates` (args: `groupId`) - fetches Object States assigned to Group with ID = `groupId`.

Also you can access object states from specific group (arg `groupId`) from the content , for example:
```graphql
{
  content {
    about(id: 2) {
      _info {
        states {
          id,
          identifier,
          priority,
          languageCodes,
          group {
            id,
            identifier,
            defaultLanguageCode,
            languageCodes,
            states {
              ...
            }
          }
        }
      }
    }
  }
}
```

# What's missing
I haven't added dynamic fields and types due to time constraints but this can be added in the future. Also mutations are not implemented yet.

### DIFF
--- a/src/GraphQL/Resolver/ObjectStateGroupResolver.php
+++ b/src/GraphQL/Resolver/ObjectStateGroupResolver.php
@@ -6,8 +6,10 @@
  */
 namespace EzSystems\EzPlatformGraphQL\GraphQL\Resolver;
 
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\ObjectStateService;
 use eZ\Publish\API\Repository\Values\ObjectState\ObjectStateGroup;
+use GraphQL\Error\UserError;
 use Overblog\GraphQLBundle\Definition\Argument;
 
 class ObjectStateGroupResolver
@@ -27,12 +29,14 @@ class ObjectStateGroupResolver
      * @param \Overblog\GraphQLBundle\Definition\Argument $args
      *
      * @return \eZ\Publish\API\Repository\Values\ObjectState\ObjectStateGroup
-     *
-     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      */
     public function resolveObjectStateGroupById(Argument $args): ObjectStateGroup
     {
-        return $this->objectStateService->loadObjectStateGroup($args['id']);
+        try {
+            return $this->objectStateService->loadObjectStateGroup($args['id']);
+        } catch (NotFoundException $e) {
+            throw new UserError("Object State Group with ID: {$args['id']} was not found.");
+        }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


webhdx:
@bdunogier I've added description to give you an overall view on what has been added. It's now ready for review.

bdunogier:
Besides the minor comments, I think that we can make this more usable:

```
_info {
  states {
    identifier
    state {
      identifier
    }
  }
}
```

`states` would yield the states groups, and `state` the state from that group assigned to the content item. Would that make sense ? Having to provide the group id at this point feels a bit odd.

webhdx:
@bdunogier I was relying on PHP API: `loadObjectStates(ObjectStateGroup $objectStateGroup, array $prioritizedLanguages = [])`. We can load all groups and run `loadObjectStates` on all of them but I'm not convinced as this might be potentially performance heavy. Up to you since it isn't a big change for me.

src/GraphQL/Resolver/ObjectStateGroupResolver.php
```diff
@@ -0,0 +1,47 @@
...
+namespace EzSystems\EzPlatformGraphQL\GraphQL\Resolver;
+
+use eZ\Publish\API\Repository\ObjectStateService;
+use eZ\Publish\API\Repository\Values\ObjectState\ObjectStateGroup;
+use Overblog\GraphQLBundle\Definition\Argument;
+
+class ObjectStateGroupResolver
+{
+    /**
+     * @var \eZ\Publish\API\Repository\ObjectStateService
```
pawbuj: phpdoc in one line

src/GraphQL/Resolver/ObjectStateResolver.php
```diff
@@ -0,0 +1,80 @@
...
+use eZ\Publish\API\Repository\ObjectStateService;
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\API\Repository\Values\ObjectState\ObjectState;
+use eZ\Publish\API\Repository\Values\ObjectState\ObjectStateGroup;
+use Overblog\GraphQLBundle\Definition\Argument;
+
+class ObjectStateResolver
+{
+    /**
+     * @var \eZ\Publish\API\Repository\ObjectStateService
```
pawbuj: phpdoc in one line

webhdx:
@bdunogier @pawbuj Fixed according to your comments. Please recheck.

webhdx:
@bdunogier Also I now load all content states without a need to specify group ID.

webhdx:
I agree. Switched to SA aware service instead: ec15e4d

mnocon:
@webhdx @bdunogier both ObjectState and ObjectStateGroup do not have the "name" property, which seems pretty important to me - shouldn't it be added as well?

PR #12, opened 2019-03-15
Remove deprecated code

Nattfarinn:
1.0 RC is best time to remove deprecations before stable release.


bdunogier:
Thank you guys. I had actually removed the routing file in the mutations PR, but it's better here :)

PR #15, opened 2019-03-18
Added description field to content types (EZP-30274)

bdunogier:
> Fixes [EZP-30274](https://jira.ez.no/browse/EZP-30274)

Adds the missing `description` field to content types:

```
{
  content {
    _types {
      article {
        _info {
          description
          french_description: description(languageCode: "fre-FR")
        }
      }
    }
  }
}
```

image

### TODO
- [x] Test on v2.5

### DIFF
--- a/src/Resources/config/graphql/ContentType.types.yml
+++ b/src/Resources/config/graphql/ContentType.types.yml
@@ -6,9 +6,14 @@ ContentType:
             id:
                 type: "Int!"
                 description: "The Content Type's unique ID."
-            #contentTypeGroups:
-            #    type: "[ContentTypeGroup]"
-            #    description: "ContentType Groups the ContentType is a member of."
+            description:
+                type: "String"
+                description: "The content type's description"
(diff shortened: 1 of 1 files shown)


bdunogier:
`"false"` looks wrong, indeed. I'll have a look, thank you @mnocon.

bdunogier:
Actually, it is a bug in `eZ\Publish\Core\Repository\Values\MultiLanguageDescriptionTrait`. It returns `false` instead of `null` when there is no description in the main language, and it is called without an argument. That's because `reset()` returns `false` when the array is empty. It is a bug in the API, is it not ?

mnocon:
I'd say yes, it doesn't make sense to me to return `false` in this case.

bdunogier:
Fixed by adding a ternary operator in the resolution expression: `resolve: "@=value.getDescription(args['languageCode']) ?: ''"`. Did the same for `name`, as it uses the same logic.

image

PR #16, opened 2019-03-19
EZP-30325: Generated schema is not being used when vendor packages are symlinked

webhdx:
Previously compiler passes were relying on relative paths to access schema definitions generated in `app/config/graphql/ezplatform` directory. This wasn't really working when symlinks were used during development. I think symlinks are still the best method when working with multiple packages so I couldn't align on my side to overcome this issue.

I have a proposal to move schema directories from constants to container parameters. It's more natural to me to store such information in the container which makes it accessible in services, extensions and compiler passes. 

I will leave it as a draft so we can validate the idea first.

### DIFF
--- a/src/DependencyInjection/Compiler/FieldValueTypesPass.php
+++ b/src/DependencyInjection/Compiler/FieldValueTypesPass.php
@@ -1,7 +1,11 @@
 <?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
 namespace EzSystems\EzPlatformGraphQL\DependencyInjection\Compiler;
 
-use EzSystems\EzPlatformGraphQL\DependencyInjection\EzSystemsEzPlatformGraphQLExtension;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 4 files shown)


webhdx:
My changes do not break test cases - it still does the same job by using `file_exists(...)` to determinate Schema configuration. 

```
➜ vendor/bin/phpspec run --format=pretty               

      EzSystems\EzPlatformGraphQL\DependencyInjection\GraphQL\YamlSchemaProvider

  23  ✔ is initializable
  28  ✔ returns the app schema if one exists
  34  ✔ returns the Domain schema if no app schema exists and the domain schema exists
  40  ✔ returns the Platform schema if no app and domain schema exist
```

PHPSpec fails in other test tho: 
```     EzSystems\EzPlatformGraphQL\Schema\Domain\Content\FieldValueBuilder\BaseFieldValueBuilder

  (...)
  50  ✘ builds ezdate
        [array:2] expected to beResolvedWithFieldValueProperty("value"), but it is
        not.
```

But it's not related to this PR (caused by 0789f087).

webhdx:
I made PR ready for review and also created a bug report on JIRA.

bdunogier:
> PHPSpec fails in other test tho:
Fixed in https://github.com/ezsystems/ezplatform-graphql/commit/c46dd65b7e0d52c019823b2dbd1d0192806ded2e today.

bdunogier:
Let's merge this :)

PR #25, opened 2019-03-26
Added plaintext output format for richtext

bdunogier:
> [EZP-30190](https://jira.ez.no/browse/EZP-30190)

Renders richtext to a plaintext version, using the HTML version passed through `strip_tags()`.

### DIFF
--- a/src/GraphQL/Resolver/RichTextResolver.php
+++ b/src/GraphQL/Resolver/RichTextResolver.php
@@ -34,4 +34,11 @@ class RichTextResolver
     {
         return $this->richTextEditConverter->convert($document)->saveHTML();
     }
+
+    public function xmlToPlainText(DOMDocument $document)
+    {
+        $html = $this->richTextConverter->convert($document)->saveHTML();
+
+        return strip_tags($html);
+    }
 }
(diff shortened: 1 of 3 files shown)


src/GraphQL/Resolver/RichTextResolver.php
```diff
@@ -34,4 +34,11 @@ public function xmlToHtml5Edit(DOMDocument $document)
     {
         return $this->richTextEditConverter->convert($document)->saveHTML();
     }
+
+    public function xmlToPlainText(DOMDocument $document)
+    {
+        $html = $this->richTextConverter->convert($document)->saveHTML();
+
+        return strip_tags($html);
```
webhdx: I'm not positive but using `$domDocument->textContent` might be better for this.

bdunogier: Oh. That is a good catch, I'll give it a try :D

bdunogier: The result looks good, but it doesn't strip HTML comments you get in debug mode.

```
the most tasty.\n  \n<!-- START vendor/ezsystems/ezpublish-kernel/eZ/Bundle/EzPublishCoreBundle/Resources/views/FieldType/RichText/embed/content.html.twig (EzPublishCoreBundle:FieldType/RichText/embed:content.html.twig) -->\n    <div class=\"align-right ez-embed-type-image\">\n    \n
```

webhdx: Alright, :+1: from me then.

lserwatka:
I guess we can merge it too 👍

PR #30, opened 2019-04-02
EZP-30414: Relation fields aren't resolved

bdunogier:
> JIRA: https://jira.ez.no/browse/EZP-30414

The `RelationFieldDefinitionMapper` is used for both `ezobjectrelation` and `ezobjectrelationlist`. The resolver is also the same, but was not correctly updated for `ezobjectrelation`, making it return null in every case for it.

The resolver has been changed to always fetch multiple items, and return one or an array depending on the multiple flag.

### DIFF
--- a/src/GraphQL/Resolver/DomainContentResolver.php
+++ b/src/GraphQL/Resolver/DomainContentResolver.php
@@ -136,13 +136,7 @@ class DomainContentResolver
 
     public function resolveDomainRelationFieldValue(Field $field, $multiple = false)
     {
-        if ($field->value instanceof FieldType\RelationList\Value) {
-            $destinationContentIds = $field->value->destinationContentIds;
-        } else if ($field->value instanceof FieldType\Relation\Value) {
-            $destinationContentIds = [$field->value->destinationContentId];
-        } else {
-            throw new UserError('\$field does not contain a RelationList or Relation Field value');
-        }
+        $destinationContentIds = $this->getContentIds($field);
 
         if (empty($destinationContentIds) || array_key_exists(0, $destinationContentIds) && is_null($destinationContentIds[0])) {
             return $multiple ? [] : null;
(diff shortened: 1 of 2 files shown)


lserwatka:
@bdunogier is this ready for QA?

src/GraphQL/Resolver/DomainContentResolver.php
```diff
@@ -136,24 +136,19 @@ public function resolveDomainFieldValue(Content $content, $fieldDefinitionIdenti
...
         } else {
-            return
-                isset($fieldValue->destinationContentIds[0])
-                    ? $this->contentLoader->findSingle(new Query\Criterion\ContentId($field->value->destinationContentIds[0]))
-                    : null;
+            throw new UserError('\$field does not contain a RelationList or Relation Field value');
         }
+
+        $contentItems = $this->contentLoader->find(new Query(
+            ['filter' => new Query\Criterion\ContentId($destinationContentIds)]
```
crosa7: Tested this with Solr enabled and it will throw an exception when $destinationContentIds is an empty array, a suggestion would be to add the following before "$contentItems = ..."  `if(empty($destinationContentIds)) {
            return $multiple ? [] : null;
        }`

bdunogier:
@lserwatka I have extracted the perf optimization commit, as it is not ready. I'll check @crosa7's feedback, and I'll mark it as ready for review (friday).

src/GraphQL/Resolver/DomainContentResolver.php
bdunogier: Thank you, very good remark @crosa7. It should be fixed now.

src/GraphQL/Resolver/DomainContentResolver.php
```diff
@@ -136,24 +136,23 @@ public function resolveDomainFieldValue(Content $content, $fieldDefinitionIdenti
 
     public function resolveDomainRelationFieldValue(Field $field, $multiple = false)
     {
-        if (!$field->value instanceof FieldType\RelationList\Value) {
-            throw new UserError("$field->fieldTypeIdentifier is not a RelationList field value");
+        if ($field->value instanceof FieldType\RelationList\Value) {
```
webhdx: Should it also work with ImageAsset fieldtype? It's also a Relation field if I'm not mistaken.

Also I'd extract the essence from this method and split into multiple methods for Relation, RelationList, ImageAssets to avoid nasty `if` statements.

bdunogier: About Image Asset, no, not really. It does load the related asset, indeed, but does extra things, and doesn't have much in common with relations overall (see `ImageAssetFieldResolver`).

I'll see what I can do to make it more readable.

bdunogier: There. I have moved the ID extraction to its own method. I could break it down into two, one for Relation, one for RelationList, but imho it is good enough that way. What do you think @webhdx ?

webhdx: It's alright :+1:

PR #31, opened 2019-04-09
Handle special characters in content type group (EZP-30422)

bdunogier:
> https://jira.ez.no/browse/EZP-30422

Changes the normalization of content types groups identifiers to transform all non alphanumeric characters to `_`. It is required on content types groups identifiers as they are actually human readable names.

### TODO
- [ ] Decide if we want to handle other types of special characters (such as accentuated, or even asian). Should we transliterate those ? If yes, using what ?

### DIFF
--- a/spec/EzSystems/EzPlatformGraphQL/Schema/Domain/Content/NameHelperSpec.php
+++ b/spec/EzSystems/EzPlatformGraphQL/Schema/Domain/Content/NameHelperSpec.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace spec\EzSystems\EzPlatformGraphQL\Schema\Domain\Content;
+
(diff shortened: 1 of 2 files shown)


bdunogier:
> have you made any decision about the TODO?

No, not yet. In any case, the fixed code is an improvement over the previous one, we don't have to decide right now I guess.

> If I create two groups: "Marek Nocoń" and "Marek-Nocoń" they are merged into one that has Content Types of both of them

That's what I call QA ! :)
The result is surprising, and I gotta admit, surprisingly good. I'd have expected one of them to be missing...

I'm not sure how we want to address that, and I probably wouldn't block this PR for that. We may have to decide at some point. The way I see it, there'll always be edge cases with the normalization we do, and the only reliable solution will in the end be to allow to customize names.

lserwatka:
Let’s call it stage 1.

PR #34, opened 2019-04-12
Added a 'viewer' field that exposes a 'user' field with the current user

bdunogier:
> https://jira.ez.no/browse/EZP-30351

Adds a `viewer` field that gives access to the current user.

```graphql
{
  viewer {
    user {
      name
      login
    }
  }
}
```

```json
{
  "data": {
    "viewer": {
      "user": {
        "name": "Administrator User",
        "login": "admin"
      }
    }
  }
}
```

### Notifications
In addition to `user`, it exposes `notifications`, that yields the viewer's notifications. Notifications data handles the specific case of content notifications, giving access to the content item.

```graphql
{
  viewer {
    notifications {
      id
      created {format(pattern: "d/m/Y H:i:s")}
      data {
        title
        message
        ... on ContentNotificationData {
          content {
            _name
            _url
          }
        }
      }
    } 
  }
}
```

```json
{
  "data": {
    "viewer": {
      "notifications": [
        {
          "id": 2,
          "created": {
            "format": "15/04/2019 16:45:53"
          },
          "data": {
            "title": "Title",
            "message": "Message",
            "content": {
              "_name": "Mexican Street Food",
              "_url": "/Media/Images/tastes/mexican-cuisine/mexican-street-food"
            }
          }
        },
        {
          "id": 1,
          "created": {
            "format": "15/04/2019 16:35:23"
          },
          "data": {
            "title": "Title",
            "message": "Message"
          }
        }
      ]
    }
  }
}
```

### DIFF
--- a/src/GraphQL/Resolver/NotificationResolver.php
+++ b/src/GraphQL/Resolver/NotificationResolver.php
@@ -1,0 +1,32 @@
+<?php
+namespace EzSystems\EzPlatformGraphQL\GraphQL\Resolver;
+
+use eZ\Publish\API\Repository\NotificationService;
+use eZ\Publish\API\Repository\Values\Notification\Notification;
+
+class NotificationResolver
+{
+    /**
+     * @var NotificationService
+     */
+    private $notificationService;
+
+    public function __construct(NotificationService $notificationService)
+    {
+        $this->notificationService = $notificationService;
+    }
+
+    public function currentUserNotifications()
+    {
+        return $this->notificationService->loadNotifications(0, 10);
+    }
+
+    public function notificationType(array $notificationData)
+    {
+        if (isset($notificationData['content_id'])) {
+            return 'ContentNotificationData';
+        } else {
+            return 'NotificationData';
+        }
+    }
+}
--- a/src/DependencyInjection/EzSystemsEzPlatformGraphQLExtension.php
+++ b/src/DependencyInjection/EzSystemsEzPlatformGraphQLExtension.php
@@ -73,7 +73,7 @@ class EzSystemsEzPlatformGraphQLExtension extends Extension implements PrependEx
     {
         $initialConfig = Yaml::parseFile(__DIR__ . '/../Resources/config/overblog_graphql.yaml');
         $initialConfig['definitions']['schema'] = (new YamlSchemaProvider($configDir))->getSchemaConfiguration();
-        $initialConfig['definitions']['types'] = ['NotificationData', 'ContentNotificationData'];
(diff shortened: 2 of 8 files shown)


src/GraphQL/Resolver/NotificationResolver.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public function currentUserNotifications()
+    {
+        return $this->notificationService->loadNotifications(0, 10);
+    }
+
+    public function notificationType(array $notificationData)
+    {
+        if (isset($notificationData['content_id'])) {
```
webhdx: It isn't really a reliable way to resolve type. I know Notifications have `$type` property and I'd expect GraphQL to use this information to resolve field type. What's more - it needs an extensibility to provide custom types for custom notification types.

bdunogier: Indeed about extensibility. It is "planned", just not done yet.

About resolving the type, well, I agree. But the values used for `$types` aren't documented anywhere :( I'll be happy to change it once I know what values are used (https://ezcrew.slack.com/conversation/C4J40K70F/p1555412078001200).

bdunogier: I have checked, and two types are used right now:
- `FlexWorkflow:Review`, in `EzSystems\FlexWorkflow\Core\SignalSlot\Slot\OnWorkflowUpdateSlot`
- `Workflow:StageChange`, in `EzSystems\EzPlatformWorkflow\Event\Subscriber\StageChangeSubscriber`

Both use the `content_id` key.
My guts would tell me to go with one notification graphQL type for each, `FlexWorkflowReviewNotificationData` and `WorkflowStageChangeNotificationData`. They would both have a `content` field. 

We could also make both of them implement a `ContentNotificationData` interface, that could be tested as in the PR's description's example to get the content. But are multiple / inherited types usable in that way, given how type resolution works ?

webhdx: For me it feels more appropriate to resolve type based on `$type` property from `Notification`. This is also a place where extensibility is needed thus we can't force particular type just because `content_id` field is available. 

I actually like your proposition to create separate types for `FlexWorkflow:Review` and `Workflow:StageChange` implementing `ContentNotificationData`.

bdunogier: @webhdx indeed, we can and should use `$type`. It would feel safer if this was more strongly typed, though.

Anyway, it implies I think that the NotificationData GraphQL types are defined by the package that uses the notification. Meaning in flex-workflow and workflow. Right ? They can't be generated since they're hardcoded in code, but they still need to be defined.

PR #39, opened 2019-06-12
EZP-30658: Upgrade `ezsystems/ezplatform-graphql` to use Symfony 4 components

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-30658


bdunogier:
No problems for me. Should we bump the version number in `composer.json` ?

mnocon:
@bdunogier this has been done already, master branch will become 2.0 (and 1.0 is, surprise, 1.0 :D )

https://github.com/ezsystems/ezplatform-graphql/blob/master/composer.json#L52

bdunogier:
Hmmm, why 2.0 ? We don't have any breaking change, do we ? I'd leave it at 1.1 until we have any.

mnocon:
For me the fact that this bundle cannot be used with kernel 7.x anymore is a breaking change of sorts (you won't be able to use it in eZ Platform 2.5, for example). That was my reasoning at least, happy to discuss this.

bdunogier:
Hmm, well, the symfony requirements will in any case prevent it from being used on v2.5...

Thinking about it, the package doesn't _require_ v7 kernel, does it ? With this change, we make it very complicated to make new features available for 2.x users (since we would need to maintain two branches at once). Most could apply there, and I think that if we can, we should give users this opportunity.

How about going with `^3.2 | ^4.3` for symfony components ?

mnocon:
I feel like this isn't really my question to answer, so I'll let you work out the best strategy with other devs 😉 From my POV we're maintaining multiple branches in almost all packages and it's not that bad.

Also, not sure I understand this part:
>  the package doesn't require v7 kernel

It has dependency on kernel package (v8 in this branch), what do you mean?

bdunogier:
>> the package doesn't require v7 kernel
> It has dependency on kernel package (v8 in this branch), what do you mean?

I meant "it doesn't require v8" kernel". Meaning that it works with it, but doesn't _need it_. As far as I know, it will work with v7 and v8, meaning that we can use both version in the _require_ line.

PR #41, opened 2019-06-13
Fixed a typo in mutations.yml

bdunogier:
Typo in `mutations.yml`, `argments` instead of `arguments`.


bdunogier:
What I know for sure is that it happens on eZ Platform 3.0 when running `bin/console`.

PR #42, opened 2019-06-17
Added content of type DomainContent to Location type

lucagentile:


bdunogier:
Thank you for your contribution @lucagentile. I'm however not sure if we want this change or not. You did see, after our discussion, that there are other, cleaner ways to achieve what you needed. I think that I'd like location to be the end of the chain, not an entry point to content.

bdunogier:
It turns out that we will add this in https://github.com/ezsystems/ezplatform-graphql/pull/51 (https://github.com/ezsystems/ezplatform-graphql/pull/51/commits/54ec4735710f16105d8f2fa2af14708b5a990e78 and https://github.com/ezsystems/ezplatform-graphql/pull/51/commits/0049c02f416077901244ce91588090f14893b740). I will close this pull-request.

Thank you again @lucagentile.

PR #43, opened 2019-06-19
Updated master to use Symfony 4.3

bdunogier:
### TODO
- [x] Figure out if this works with richtext (moved to external package between 2.x and 3.x
- [x] See what should be done in regards to `.yml` vs `.yaml`
- [x] Rebase on top of `master` once #47 and #46 are merged to `1.0` and `master`. 
- [ ] Once added to the meta, make sure to remove the default schema config for `query` added by the recipe to `config/packages/graphql.yaml`

### DIFF
--- a/src/DependencyInjection/EzSystemsEzPlatformGraphQLExtension.php
+++ b/src/DependencyInjection/EzSystemsEzPlatformGraphQLExtension.php
@@ -20,11 +20,11 @@ use Symfony\Component\DependencyInjection\Loader;
  */
 class EzSystemsEzPlatformGraphQLExtension extends Extension implements PrependExtensionInterface
 {
-    private const APP_SCHEMA_DIR_RELATIVE_PATH = '/config/graphql';
-    private const EZPLATFORM_SCHEMA_DIR_RELATIVE_PATH = '/ezplatform';
-    private const PACKAGE_DIR_RELATIVE_PATH = '/../vendor/ezsystems/ezplatform-graphql';
-    private const PACKAGE_SCHEMA_DIR_RELATIVE_PATH = '/src/Resources/config/graphql';
-    private const FIELS_DEFINITION_FILE_NAME = 'Field.types.yml';
+    private const SCHEMA_DIR_PATH = '/config/graphql/types';
+    private const EZPLATFORM_SCHEMA_DIR_PATH = '/ezplatform';
+    private const PACKAGE_DIR_PATH = '/vendor/ezsystems/ezplatform-graphql';
+    private const PACKAGE_SCHEMA_DIR_PATH = '/src/Resources/config/graphql';
+    private const FIELDS_DEFINITION_FILE_NAME = 'Field.types.yaml';
 
     /**
      * {@inheritdoc}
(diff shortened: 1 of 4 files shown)


bdunogier:
One issue we may have though is the richtext part, since the namespace for some items we use has changed. We should be able to work it out with some dynamically loaded service definition, though.

webhdx:
I don't think this is worth the maintenance hell. We intentionally avoided to do so in other packages. Realistically we are not going to (or we shouldn't given we use semver) add more stuff to GraphQL. Also keep in mind that with Symfony 3.4 we didn't use Flex directory structure. I doubt we can easily handle flex/non flex application in generate schema scripts as it heavily relies on hardcoded paths.

src/DependencyInjection/EzSystemsEzPlatformGraphQLExtension.php
```diff
@@ -20,11 +21,15 @@
...
-    private const FIELS_DEFINITION_FILE_NAME = 'Field.types.yml';
+    /**
+     * Symfony 3 backward compatibility
+     */
+    private const APP_DIR_PATH = '/app';
+    private const SCHEMA_DIR_PATH = '/config/graphql';
+    private const EZPLATFORM_SCHEMA_DIR_PATH = '/ezplatform';
+    private const PACKAGE_DIR_PATH = '/vendor/ezsystems/ezplatform-graphql';
+    private const PACKAGE_SCHEMA_DIR_PATH = '/src/Resources/config/graphql';
+    private const FIELDS_DEFINITION_FILE_NAME = 'Field.types.yml';
```
webhdx: `Field.types.yaml`?

bdunogier:
I'm actually close to having the path thing working with both Sf 3.x and 4.x, if you're willing to accept the change.

In my opinion, it is worth trying to release a 1.1, and maybe a 1.2, that are compatible with 2.5 LTS. I'll push the changes later today, and you can review them. If you think it's really not worth it, we can still go for a 2.0. But we should really not underestimate the expectations from 2.x users.

src/DependencyInjection/EzSystemsEzPlatformGraphQLExtension.php
bdunogier: Indeed, thank you.

bdunogier:
It was getting too complicated with some aspects, so I have made it Sf 4 only. With the changes documented in the TODO for the meta, it works on the current ezplatform master for me.

bdunogier:
The only change should be `Updated paths for Symfony 4`. The other commits are merged into `1.0` and `master`.

webhdx:
@bdunogier Looks like naming scheme is different for Flex applications: https://github.com/overblog/GraphQLBundle/blob/master/docs/definitions/schema.md#default-files-location
They ditched `*.types.yml` in favor of `types/*.yaml`. Do you think we should also align our schema to follow this scheme?

bdunogier:
Good catch, @webhdx. I think we should adopt it. But could we merge this one, and then issue another PR for this change ? I'd like to have a version that works with ezp 3.x asap :)

PR #44, opened 2019-06-24
Bumped php-cs-fixer to v2.15 and added CS check job

alongosz:
|      |     |
| --- | --- |
| **Type** | Bug |
| **Target version** | `v1.0.2` for eZ Platform `v2.5.2 LTS` |

This PR:
- [x] bumps php-cs-fixer to v2.15.1 (locking `>=2.15.1 <2.16.0`),
- [x] adds composer cache directory to Travis cached directories,
- [x] adds a Travis job which checks code style ([failing build](https://travis-ci.org/ezsystems/ezplatform-rest/jobs/549685935), [passing build](https://travis-ci.org/ezsystems/ezplatform-graphql/builds/549706248?utm_source=github_status&utm_medium=notification)),
- [x] adds running Travis on stable branches (e.g. `1.0`),
- [x] disables unwanted CS rules,
- [x] fixes outstanding CS issues.

PR #45, opened 2019-06-24
EZP-30691: Dropped unused dependency from GeneratePlatformSchemaCommand

alongosz:
|      |     |
| --- | --- |
| **Type** | Bug |
| **Target version** | `v1.0.2` for eZ Platform `v2.5.2 LTS` |
| **Requires** | #44 

Commit relevant for review: ef19288

The `ezplatform:graphql:generate-schema `(`\EzSystems\EzPlatformGraphQL\Command\GeneratePlatformSchemaCommand`) command relies on Repository but does not use it.

Right now it only generates warnings (introduced via ezsystems/ezpublish-kernel#2636) from time to time, when building container, so we should drop it.

**TODO**
- [x] **Before merge**: rebase after merging #44

### DIFF
--- a/src/Command/GeneratePlatformSchemaCommand.php
+++ b/src/Command/GeneratePlatformSchemaCommand.php
@@ -7,7 +7,6 @@
 namespace EzSystems\EzPlatformGraphQL\Command;
 
 use EzSystems\EzPlatformGraphQL\Schema\Generator;
-use eZ\Publish\API\Repository\Repository;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


alongosz:
> +1, but as said on slack, curious why this triggers load as repo service should lazy, but I guess this using autowiring triggers siteaccess aware repo which is not lazy atm? _(I guess it should be, but for 3.0 only the "tip" of repo should be lazy)_

Hmm, actually, right now I cannot reproduce warning coming from that command. What was auto-wired was `eZ\Publish\Core\SignalSlot\Repository`, so I might be wrong blaming Repository here. As we know - those are hard to debug ;)

bdunogier:
Note that I do get the warning about the argument with Sf 4 (if I understand the conversation above).

alongosz:
> Actually, for the sake of BC, we should keep the argument and make it optional. And make the Command a `final` class in 2.0 :)

@bdunogier there's no BC for constructors, unless a class is a part of an SPI. Moreover this is 1st argument, with 2nd being not optional, so what you're suggesting is technically not possible :)

+1 for `final` in 2.0 regardless.

bdunogier:
Okay :)

bdunogier:
Well, if somebody extended the constructor (spi or not, it is possible because no final), and calls the parent's, then it will fatal -> backward compatibility was broken. But it's not important enough to spend energy on this, it is too hot 🌞

PR #47, opened 2019-06-26
Changed the generation command to pick the path from the container parameters

bdunogier:
No reason why it should not.

### DIFF
--- a/src/Resources/config/services/services.yml
+++ b/src/Resources/config/services/services.yml
@@ -4,8 +4,9 @@ services:
         autowire: true
         public: false
 
-    EzSystems\EzPlatformGraphQL\Command\:
-        resource: '../../Command/*'
+    EzSystems\EzPlatformGraphQL\Command\GeneratePlatformSchemaCommand:
+        arguments:
+            $schemaRootDir: '%ezplatform.graphql.schema.root_dir%'
         tags:
             -  { name: console.command }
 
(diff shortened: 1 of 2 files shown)


src/Resources/config/services/services.yml
```diff
@@ -4,8 +4,9 @@ services:
         autowire: true
         public: false
 
-    EzSystems\EzPlatformGraphQL\Command\:
-        resource: '../../Command/*'
+    EzSystems\EzPlatformGraphQL\Command\GeneratePlatformSchemaCommand:
+        arguments:
+            $schemaRootDir: '%ezplatform.graphql.schema.root_dir%'
```
alongosz: Side:
Is it just me or is this parameter [set](https://github.com/ezsystems/ezplatform-graphql/blob/v1.0.1/src/DependencyInjection/EzSystemsEzPlatformGraphQLExtension.php#L86) [twice](https://github.com/ezsystems/ezplatform-graphql/blob/v1.0.1/src/DependencyInjection/Compiler/RegisterSchemaDirectoryParametersPass.php#L28)? :)

bdunogier: Hummm, indeed. Leftovers from the initial implementation I guess. I'll get rid of the compiler pass, I think.

bdunogier:
All fixed besides the compiler pass, thank you for the review @alongosz.

PR #51, opened 2019-07-05
Location by id with children support

bdunogier:
> JIRA: https://jira.ez.no/browse/EZP-30726
> Target version: 1.1 (merges to master for 2.0 release without any issue)

- Adds `Repository.location: Location` field that returns a location by id.
- Adds `Location.children` that returns a connection with the location's children
- Adds `Location.content: Content`
- Adds `ContentInfo.currentVersion: Version`
- Adds `Version.creator: User`
- Adds `Content._thumbnail: Thumbnail`

Also adds page number based pagination support using a new connection field, `LocationConnection.pages`.

Example:
```
{
  _repository {
    location(locationId: 2) {
      pathString
      children {
        pages {
          number
          cursor
        }
        edges {
          node {
            content {
              name
              currentVersion {
                creator {
                  name
                }
              }
            }
          }
        }
      }
    }
  }
}
```

### TODO
- [x] Define a dedicated alias for thumbnails (200x200 pixels)
- [x] Change `Location.id` to be an `ID`, and add `Location.locationId` with the auto-increment id.

### DIFF
--- a/src/GraphQL/DataLoader/SearchLocationLoader.php
+++ b/src/GraphQL/DataLoader/SearchLocationLoader.php
@@ -6,61 +6,60 @@
  */
 namespace EzSystems\EzPlatformGraphQL\GraphQL\DataLoader;
 
+use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\Values\Content\LocationQuery;
 use EzSystems\EzPlatformGraphQL\GraphQL\DataLoader\Exception\ArgumentsException;
 use eZ\Publish\API\Repository\Exceptions as ApiException;
 use eZ\Publish\API\Repository\SearchService;
-use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\Content\Query;
 use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
-use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
 
 /**
  * @internal
  */
-class SearchContentLoader implements ContentLoader
+class SearchLocationLoader implements LocationLoader
 {
     /**
      * @var SearchService
      */
     private $searchService;
 
-    public function __construct(SearchService $searchService)
+    /**
+     * @var LocationService
+     */
+    private $locationService;
+
+    public function __construct(SearchService $searchService, LocationService $locationService)
     {
         $this->searchService = $searchService;
+        $this->locationService = $locationService;
     }
 
-    /**
-     * Loads a list of content items given a Query Criterion.
-     *
-     * @param Query $query A Query Criterion. To use multiple criteria, group them with a LogicalAnd.
-     *
-     * @return \eZ\Publish\API\Repository\Values\Content\Content[]
-     *
-     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
-     */
-    public function find(Query $query): array
+    public function find(LocationQuery $query): array
     {
         return array_map(
             function (SearchHit $searchHit) {
                 return $searchHit->valueObject;
             },
-            $this->searchService->findContent($query)->searchHits
+            $this->searchService->findLocations($query)->searchHits
         );
     }
 
-    /**
-     * Loads a single content item given a Query Criterion.
-     *
-     * @param Criterion $filter A Query Criterion. Use Criterion\ContentId, Criterion\RemoteId or Criterion\LocationId for basic loading.
-     *
-     * @return \eZ\Publish\API\Repository\Values\Content\Content
-     *
-     * @throws ArgumentsException
-     */
-    public function findSingle(Criterion $filter): Content
+    public function findById($id): Location
     {
         try {
-            return $this->searchService->findSingle($filter);
+            return $this->locationService->loadLocation($id);
+        } catch (ApiException\InvalidArgumentException $e) {
+        } catch (ApiException\NotFoundException $e) {
+            throw new ArgumentsException($e->getMessage(), $e->getCode(), $e);
+        }
+    }
+
+    public function findByRemoteId($id): Location
+    {
+        try {
+            return $this->locationService->loadLocationByRemoteId($id);
         } catch (ApiException\InvalidArgumentException $e) {
         } catch (ApiException\NotFoundException $e) {
             throw new ArgumentsException($e->getMessage(), $e->getCode(), $e);
@@ -76,14 +75,14 @@ class SearchContentLoader implements ContentLoader
      *
      * @throws ArgumentsException
      */
-    public function count(Query $query)
+    public function count(LocationQuery $query)
     {
         $countQuery = clone $query;
         $countQuery->limit = 0;
         $countQuery->offset = 0;
 
         try {
-            return $this->searchService->findContent($countQuery)->totalCount;
+            return $this->searchService->findLocations($countQuery)->totalCount;
         } catch (ApiException\InvalidArgumentException $e) {
             throw new ArgumentsException($e->getMessage(), $e->getCode(), $e);
         }
--- a/src/GraphQL/InputMapper/SearchQuerySortByMapper.php
+++ b/src/GraphQL/InputMapper/SearchQuerySortByMapper.php
@@ -1,0 +1,50 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformGraphQL\GraphQL\InputMapper;
+
+use eZ\Publish\API\Repository\Values\Content\Query;
+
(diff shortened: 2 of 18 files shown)


DubMan21:
Hello, it's a good improvement ! 
But some features would be interesting : being able to query parents and children as on the master branch (content { TYPE(_query: {....}_) {} }).

Especially query by Depth and by ContentType

bdunogier:
> But some features would be interesting, being able to query parents and children as on the master branch : "content { TYPE(query: {....}) {} }" .

> Especially query by Depth and by ContentType

Could you please elaborate a bit ? I'm not sure what you mean here. It is already possible to query a type of item.

DubMan21:
I currently use "_repository" and "children" to get my content structure tree from eZ Platform with this query  :

```
{
  _repository {
    location(locationId: 2 ) {
      children {
        edges {
          node {
            urlAliases {
              path 
... }
```

But I get all the Content Type and I would like to filter by one specific content type. We could have a query like this :

```
{
  _repository {
    location(locationId: 2 ) {
      children**(query: {ContentType: })** {
        edges {
          node {
            urlAliases {
              path 
... }
```

And for the depth, we could have a parameter for choosing the maximum depth of the query (e.g. fetching the first two levels of children of the content).

bdunogier:
I see, it makes sense. While it should be fairly easy to implement, I have an alternative solution to that use-case.

Usually, when a content's sub-items of a particular type are retrieved, they have a meaning in terms of content. For instance, the `blog_post` sub-items of a `blog` are the posts of that blog. A gallery's image sub-items are the gallery's images. And so on.

I have designed a fieldtype, https://github.com/bdunogier/ezplatform-query-fieldtype, that models exactly this. You can create `gallery.images`, or `blog.posts`, and query them, be it with PHP, REST or GraphQL:

```
{
  content {
    gallery {
      name
      images {
        name
        image {
          variation(identifier: thumbnail) { url alternativeText }
        }
      }
    }
  }
}
```

The field definition is configured using:
- a returned type (example: images)
- a QueryType (example: children of a location of a given type)
- a mapping of the QueryType's parameters to either static values, or properties / fields from the current content (example: the gallery item's main location id)

I strongly believe that it is a cleaner, more maintainable solution to the use-case you describe. It lets the content architect model the relationships that exist between items, and lets content consumers focus on rendering the content.

I'm available on slack if you need more information.

bdunogier:
Ping @webhdx for review :-)

src/GraphQL/InputMapper/SearchQuerySortByMapper.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    /**
+     * @param array $inputArray
+     *
+     * @return \eZ\Publish\API\Repository\Values\URL\Query\SortClause[]
+     */
+    public function mapInputToSortClauses(array $sortInput)
+    {
+        $sortClauses = array_map(
+            function ($sortClauseClass) {
```
webhdx: Strong typing would be helpful here

src/GraphQL/InputMapper/SearchQuerySortByMapper.php
```diff
@@ -0,0 +1,50 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformGraphQL\GraphQL\InputMapper;
+
+use eZ\Publish\API\Repository\Values\Content\Query;
+
+class SearchQuerySortByMapper
+{
+    /**
+     * @param array $inputArray
```
webhdx:
```suggestion
     * @param string[] $sortInput
```

src/GraphQL/Relay/Page.php
```diff
@@ -0,0 +1,29 @@
...
+class Page
+{
+    public function __construct(int $number, string $cursor)
+    {
+        $this->number = $number;
+        $this->cursor = $cursor;
+    }
+
+    /**
+     * @var int
```
webhdx: Single line docblock, constructor should be placed under class properties

src/GraphQL/Relay/PageAwareConnection.php
```diff
@@ -0,0 +1,51 @@
...
+    /** @var Edge[] */
+    public $edges = [];
+
+    /** @var PageInfo */
+    public $pageInfo;
+
+    /** @var int */
+    public $totalCount;
+
+    /** @var Page */
```
webhdx:
```suggestion
    /** @var Page[] */
``` 
???

src/GraphQL/Resolver/ContentThumbnailResolver.php
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * @param \eZ\Publish\API\Repository\Values\Content\Content $content
+     *
+     * @return Field
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    private function getThumbnailImageField(Content $content)
```
webhdx: Return type

src/GraphQL/Resolver/DomainContentResolver.php
```diff
@@ -72,12 +72,15 @@ public function resolveDomainContentItems($contentTypeIdentifier, $query = null)
...
+     * @param \Overblog\GraphQLBundle\Definition\Argument|array $args
+     * @param string|null $contentTypeIdentifier
      *
      * @return \eZ\Publish\API\Repository\Values\Content\Content
+     *
+     * @throws \GraphQL\Error\UserError if $contentTypeIdentifier was specified, and the loaded item's type didn't match it
+     * @throws \GraphQL\Error\UserError if no argument was provided
      */
-    public function resolveDomainContentItem(Argument $args, $contentTypeIdentifier)
+    public function resolveDomainContentItem($args, $contentTypeIdentifier)
```
webhdx: Did it change in overblog graphql package? I though it would always return an `Argument` object which implements `ArrayAccess` if I'm not mistaken.

src/GraphQL/DataLoader/SearchLocationLoader.php
```diff
@@ -0,0 +1,90 @@
...
+        );
+    }
+
+    public function findById($id): Location
+    {
+        try {
+            return $this->locationService->loadLocation($id);
+        } catch (ApiException\InvalidArgumentException $e) {
+        } catch (ApiException\NotFoundException $e) {
+            throw new ArgumentsException($e->getMessage(), $e->getCode(), $e);
```
webhdx: There happens something confusing to me. Why does `ArgumentsException` extend `NotFoundException` if it thrown for `NotFoundException` and for `InvalidArgumentException` below? To my understanding it should not extend any particular Exception class because it's very general exception. Also do we need specific `ArgumentsException` class if we have `InvalidArgumentException` from our API?

src/GraphQL/Resolver/DomainContentResolver.php
bdunogier: I did change this because in this PR, to resolve a Location's `content` field, we call the method directly, with an array as the argument. In `src/Resources/config/graphql/Location.types.yml`: `resolve: "@=resolver('DomainContentItem', [{id: value.contentId}, null])"`.

I agree that it's not the cleanest thing ever, but... 😊

src/GraphQL/DataLoader/SearchLocationLoader.php
bdunogier: The general idea is to make that package only throw its own exceptions (it supposedly catches everything else). Why it extends `NotFoundException`... I don't know. A mistake I guess. I'll change it extend `\Exception`. Would that work for you ?

Generally speaking, error handling in GraphQL needs to be vastly improved, but I have barely scratched the surface of it.

bdunogier:
Review comments taken care of, except the two I have answered about, @webhdx.
I'm still thinking about the exception topic. Not as easy as it sounds in the end... (ref. the tough error handling topic in the comment above).

src/GraphQL/DataLoader/SearchLocationLoader.php
bdunogier: After thinking about it, I'd like to request your approval to leave this aside for now. The reason why it is implemented that way in this loader is that it was implemented that way in the existing ones.

We must find a proper way to implement exception handling in loaders, but I don't think that we must do it in the context of that PR.

bdunogier:
Also, the PR has been rebased to contain coherent commits.

bdunogier:
@ezsystems/qa-team, the baby is yours.

micszo:
It would be helpful to document that changes to these items affect the schema (and require regenerating it): content types, types groups, languages, blocks types, images variations. 
Also we need to make sure the list is complete. (:

bdunogier:
@micszo I'll rebase the PR to squash the fixup commit.

micszo:
@lserwatka this one's ready

lserwatka:
@bdunogier you can merge it up

lserwatka:
Merged via https://github.com/ezsystems/ezplatform-graphql/commit/219501b1ca4f9eed8827f6526323b580d09ec6b3

PR #53, opened 2019-08-20
[WIP] Install latest dependencies

mnocon:


bdunogier:
Could you explain why these requirements were necessary, @mnocon ? This package doesn't, afaik, depend on:
- doctrine-dbal-schema (it doesn't manipulate any DB schema)
- repository forms (no forms handling)
- ezplatform-admin-ui-modules
- ezplatform-rest
- ez-support-tools
- ezplatform-user

mnocon:
Hi @bdunogier! Please see discussion in https://github.com/ezsystems/ezplatform-http-cache/pull/99#pullrequestreview-277760933 , as this is the same situation (additional requirements come from dependency on kernel and AdminUI)

Long story short, after you remove `prefer-stable, minimum stability` the package fails to install (as in: you clone it and perform `composer install`) until you add those packages.

But this PR is still WIP, as I'm not sure what's the best way to do this (Andre suggested going with require-dev, which I like more and more)

bdunogier:
I see, thank you for the explanation.

If the use-case is indeed to install the requirements for CI / tests, `require-dev` feels like a clean way out.

PR #55, opened 2019-09-20
Fix EZP-30957: avoid exceptions on content types not in the generated schema

bdunogier:
After a content type has been added, but before the schema has been generated, trying to list
a location's sub-items that contain one of those items will error out. The typeResolver of the
DomainContent interface will return for items of those types a name that doesn't exist in the schema.

To work around that, content items of types that don't exist in the schema are mapped to an "anonymous" type, "UntypedContent". It has the properties shared by all content items (`_url`, `_name`, `thumb` etc, but not the fields.

### TODO
- [x] ~~Consider renaming the type. `AnonymousContent`~~ (done, to `UntypedContent`)?
- [x] Rebase against the 1.x branch as it is also affected

### DIFF
--- a/src/DependencyInjection/GraphQL/YamlSchemaProvider.php
+++ b/src/DependencyInjection/GraphQL/YamlSchemaProvider.php
@@ -37,6 +37,7 @@ class YamlSchemaProvider implements SchemaProvider
             'query' => $this->getQuerySchema(),
             'mutation' => $this->getMutationSchema(),
             'resolver_maps' => [UploadMap::class],
+            'types' => ['UntypedContent'],
         ];
     }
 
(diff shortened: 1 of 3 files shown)


mnocon:
@bdunogier this PR is now failing for me (it worked correctly before the name change), could it be because `UngeneratedContent` is still mentioned in the code?

bdunogier:
> could it be because UngeneratedContent is still mentioned in the code?

Yes, probably. I'll fix it right away.

bdunogier:
It should be fixed now @mnocon.

lserwatka:
@bdunogier could you merge it up?

mnocon:
Merged to master in https://github.com/ezsystems/ezplatform-graphql/commit/46a40328a8bbf4ee1c069ac069aab021f889d604

PR #56, opened 2019-09-21
Added a language argument on content fields

bdunogier:
Gives more flexibility when querying multilingual content. As long as the current siteaccess is configured with several (or all) languages, multiple translations of each field can be obtained at once.

image

### DIFF
--- a/src/GraphQL/Resolver/DomainContentResolver.php
+++ b/src/GraphQL/Resolver/DomainContentResolver.php
@@ -132,9 +132,9 @@ class DomainContentResolver
         return isset($aliases[0]->path) ? $aliases[0]->path : null;
     }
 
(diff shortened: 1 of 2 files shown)


nague:
Hi @bdunogier  ! 

We are currently on a project and this feature would be useful for us.
Is it possible to know when it could be finalized ?

Thanks a lot for your work.

bdunogier:
Hi @nague. Thank you very much for your feedback. We don't really have a date for the feature yet, to be honest. We came up with the idea a couple weeks ago, and I wanted to see if it would raise interest. It looks like it does :-)

We will report your feedback in our internal tool. Do you mind sharing more details about your project so we can be more accurate in the report and inform you when this feature will be released. You can do so by email if you prefer, bd at ez dot no.

PR #58, opened 2019-09-26
[Fix] Keeping the backoffice order in the GraphQL result of the relation list

jevillard:
## This PR is a fix for relationlist.
In the backoffice, I have a relationlist with a specific order. I hope to recover my relationlist with the same order in the result of my GraphQL query.

### Before PR
ez-relationlist-graphql

### After PR
ez-relationlist-graphql-after-pr

### DIFF
--- a/src/GraphQL/Resolver/DomainContentResolver.php
+++ b/src/GraphQL/Resolver/DomainContentResolver.php
@@ -149,7 +149,16 @@ class DomainContentResolver
             ['filter' => new Query\Criterion\ContentId($destinationContentIds)]
         ));
 
-        return $multiple ? $contentItems : $contentItems[0] ?? null;
+        if ($multiple) {
+            return array_map(
+                function ($contentId) use ($contentItems) {
+                    return $contentItems[array_search($contentId, array_column($contentItems, 'id'))];
+                },
+                $destinationContentIds
+            );
+        }
+
+        return $contentItems[0] ?? null;
     }
 
     public function resolveDomainContentType(Content $content)


src/GraphQL/Resolver/DomainContentResolver.php
```diff
@@ -149,7 +149,17 @@ public function resolveDomainRelationFieldValue(Field $field, $multiple = false)
             ['filter' => new Query\Criterion\ContentId($destinationContentIds)]
         ));
 
-        return $multiple ? $contentItems : $contentItems[0] ?? null;
+        if ($multiple) {
+            $destinationContentItems = [];
```
bdunogier: You could shorten this to:

```
return array_map(
  function ($contentId) use ($contentItems) {
    $key = array_search($contentId, array_column($contentItems, 'id'));
    return $contentItems[$key];
  }
```

I love `array_map()` :-)

lserwatka:
@jevillard are you able to come up with updated PR by Monday? We are tagging eZ Platform 3.0.0-beta2 late Monday next week. We would like to include your fix.

jevillard:
@bdunogier `php_cs` add a blank line before the return statement in array_map.

To avoid having this line do you prefer that I write the return in this way or not?
```return $contentItems[array_search($contentId, array_column($contentItems, 'id'))];```

In any case, the PR is ready and rebase on 1.0

bdunogier:
> To avoid having this line do you prefer that I write the return in this way or not?
return $contentItems[array_search($contentId, array_column($contentItems, 'id'))];

Why not. It's a bit compact, but still acceptable I think.

lserwatka:
@bdunogier could you merge it up?

bdunogier:
Done @lserwatka.

PR #61, opened 2019-10-07
Added the FieldDefinitionInputMapper interface

bdunogier:
> Required by the GraphQL implementation of https://github.com/ezsystems/ezplatform-matrix-fieldtype/pull/20

Implement on a Field Definition Mapper to generate a dynamic type for create and update mutations
on a given Field Type. The service must be tagged with `ezplatform_graphql.field_definition_input_mapper`, with a fieldtype attribute set to the Field Type's identifier.

In version 2.0, the new interface will be merged into `FieldDefinitionMapper`, and the service tag won't be needed anymore.


bdunogier:
This should be ready for review, @webhdx.

doc/howto/custom_field_type.md
```diff
@@ -160,6 +160,31 @@ class RelationFieldDefinitionMapper extends DecoratingFieldDefinitionMapper impl
...
+generates its own input types depending on the configured columns.  It defines an extra method, `mapToFieldValueInputType`, 
+that returns a GraphQL type for a Field Definition.
+
+Example:
+```
+class MyFieldDefinitionMapper extends BaseFieldDefinitionMapper implements FieldDefinitionMapper, FieldDefinitionInputMapper
+{
+    public function mapToFieldValueInputType(ContentType $contentType, FieldDefinition $fieldDefinition): ?string
+    {
+        // Forward compatibility with 2.x.
```
alongosz: I'd rather say either `// Forward compatibility with eZ Platform 3.x.` or `// Forward compatibility with version 2.x` of this bundle.
Because those versions are close with eZ Platform, it might be confusing.

doc/howto/custom_field_type.md
```diff
@@ -160,6 +160,31 @@ class RelationFieldDefinitionMapper extends DecoratingFieldDefinitionMapper impl
...
+        // It will only call the parent's method if it exists.
+        if (!$this->canMap($fieldDefinition) && is_callable('parent::mapToFieldValueInputType')) {
```
alongosz: Interesting forward compatibility approach :)

In this case I'd rather go for implementing `mapToFieldValueInputType` in `BaseFieldDefinitionMapper` directly, so you can skip ` && is_callable('parent::mapToFieldValueInputType')` in this example.

But don't have a strong opinion here :)

src/Schema/Domain/Content/Mapper/FieldDefinition/FieldDefinitionInputMapper.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition;
```
alongosz:
```suggestion
declare(strict_types=1);

namespace EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition;
```

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
```diff
@@ -95,12 +101,21 @@ protected function getFieldDefinitionField(FieldDefinition $fieldDefinition): st
         return $this->getNameHelper()->fieldDefinitionField($fieldDefinition);
     }
 
-    private function getFieldDefinitionToGraphQLType(FieldDefinition $fieldDefinition, $operation): string
+    private function fieldType(array $args, $operation): string
```
alongosz: naming: this should still start with `get`  or something similar (I guess it's not a field type identifier, but identifier + string describing something, so the name should also reflect that.

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
```diff
@@ -95,12 +101,21 @@ protected function getFieldDefinitionField(FieldDefinition $fieldDefinition): st
...
     {
+        $fieldDefinition = $args['FieldDefinition'];
+        $contentType = $args['ContentType'];
```
alongosz: Since you've already extracted this in `work` method, maybe here you could pass them directly instead of reading `$args` again?
I mean passing:
`ContentType $contentType, FieldDefinition $fieldDefinition` as arguments to this method.

doc/howto/custom_field_type.md
bdunogier: I like the idea. This check was acceptable for forward compatibility I think, but the abstract implementation seems to make more sense. I'll give it a try, thank you !

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
bdunogier: Yeah, but... it's a standard I've been following in the Workers. There is a lot of name building, that repeats a lot, and you end up with a looot of `get` calls.  See https://github.com/ezsystems/ezplatform-graphql/blob/master/src/Schema/Domain/Content/Worker/ContentType/AddContentOfTypeConnectionToDomainGroup.php#L57-L75 for an example.

I understand the remark, and to be honest, thought about the same, but I chose to favour consistency here.

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
bdunogier: You're right, good catch.

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
bdunogier: On second thought, I'll unify them with "get" in that package.

doc/howto/custom_field_type.md
bdunogier: Done in dec6ddf4ac5c3eb7397f67cfc34ba9e2ea620be4.

doc/howto/custom_field_type.md
bdunogier: Fixed.

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
bdunogier: Changed naming methods in that class. They now use the verb "name".

bdunogier:
Ping @webhdx if you could have a look please. It is required by the matrix implementation, and I need to update it for master after merging :)

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
```diff
@@ -62,45 +65,54 @@ public function canWork(Builder $schema, array $args): bool
...
     }
 
     /**
-     * @param ContentType $contentType
+     * @param array $args
      *
      * @return string
      */
-    protected function getCreateInputName(ContentType $contentType): string
+    protected function nameCreateInputType(array $args): string
```
alongosz: I still don't follow why you prefer passing array of not well-known objects instead of strictly-typed `ContentType` for this method. Am I missing the bigger picture here?

src/Schema/Domain/Content/Mapper/FieldDefinition/DecoratingFieldDefinitionMapper.php
```diff
@@ -35,6 +37,16 @@ public function mapToFieldValueResolver(FieldDefinition $fieldDefinition): ?stri
...
     }
 
+    public function mapToFieldValueInputType(ContentType $contentType, FieldDefinition $fieldDefinition): ?string
+    {
+        /** @deprecated this test will be removed in ezplatform-graphql 2.x */
+        if ($this->innerMapper instanceof FieldDefinitionInputMapper) {
+            return $this->innerMapper->mapToFieldValueInputType($contentType, $fieldDefinition);
+        }
+
+        throw new LogicException('The inner mapper is not a FieldDefinitionInputMapper. This method should not have been called in an 1.x versions of ezplatform-graphql.');
```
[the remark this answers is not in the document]
bdunogier: In "an 1.x" ? It does sounds wrong, indeed. I'll change it to "the".

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
bdunogier: Because these methods are called over and over in the worker classes. It increases readability a lot. I actually started with explicit arguments, but decided to change it when I saw it grow in complexity.

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
```diff
@@ -24,33 +24,40 @@ class AddFieldDefinitionToDomainContentMutation extends BaseWorker implements Wo
...
+        $this->mappers = $mappers;
     }
 
     public function work(Builder $schema, array $args)
     {
-        $fieldDefinition = $args['FieldDefinition'];
-        $contentType = $args['ContentType'];
-
-        $fieldDefinitionField = $this->getFieldDefinitionField($fieldDefinition);
+        $properties = ['description' => $args['FieldDefinition']->getDescriptions()['eng-GB'] ?? ''];
```
webhdx: You can't really expect that `eng-GB` translation will be available. We don't force any particular naming convention for language identifiers. Consider:
* using main language from Content Type
* using first element from `getDescriptions()` array

bdunogier: Yeah... well, if you look at the rest of the code, this issue exists in many places :)

But for this one, I'll change it to use the main language, good tip, thank you.

src/Schema/Domain/Content/Mapper/FieldDefinition/DecoratingFieldDefinitionMapper.php
DominikaK: How about:
```suggestion
        throw new LogicException('The inner mapper is not a FieldDefinitionInputMapper. This method should not be called in ezplatform-graphql 1.x.');
```

bdunogier: The passive phrasing was because it is not called by the code user, but by the code itself. A developer can't call that directly.

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldDefinitionToDomainContentMutation.php
bdunogier: d8f61b535f590cdebe127dd43f96e1d5c2027658. Better, @webhdx ?

bdunogier:
Rebased, and fixed the sentence discussed above. I'll merge to 1.0, and proceed with the updates for master (please, do NOT merge 1.0 into master).

PR #66, opened 2019-10-22
Add full query support for ezxmltext field type

dfritschy:
This PR provides full query support for `ezxmltext` field types.

Modelled after the richtext field type, it offers these subtypes:

          text:
                type: "String"
                description: "String representation of the value"
            xml:
                type: "String"
                description: "The raw ezxmltext xml"
            plaintext:
                type: "String"
                description: "Plain text representation of the value, without tags. Warning: the text representation may not be perfect."
            html5:
                type: "String"
                description: "HTML5 representation."

Note: mutations are not in the scope of this PR.

### DIFF
--- a/src/Resources/config/graphql/Field.types.yaml
+++ b/src/Resources/config/graphql/Field.types.yaml
@@ -260,3 +260,24 @@ SelectionFieldValue:
                 type: "String"
                 description: "String representation of the value"
                 resolve: "@=value"
+
+XmlTextFieldValue:
+    type: object
+    config:
+        fields:
+            text:
+                type: "String"
+                description: "String representation of the value"
+                resolve: "@=value"
+            xml:
+                type: "String"
+                description: "The raw ezxmltext xml"
+                resolve: "@=value"
+            plaintext:
+                type: "String"
+                description: "Plain text representation of the value, without tags. Warning: the text representation may not be perfect."
+                resolve: "@=value.xml.textContent"
+            html5:
+                type: "String"
+                description: "HTML5 representation."
+                resolve: "@=resolver('XmlTextToHtml5', [value.xml])"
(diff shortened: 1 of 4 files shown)


src/Resources/config/graphql/Field.types.yaml
```diff
@@ -260,3 +260,24 @@ SelectionFieldValue:
...
                 resolve: "@=value"
+
+XmlTextFieldValue:
+    type: object
+    config:
+        fields:
+            text:
+                type: "String"
+                description: "String representation of the value"
+                resolve: "@=value"
```
bdunogier: You can remove that line. As far as I know it is the default.

dfritschy:
Thanks, @bdunogier for your comments.

> it should be against the 1.0 branch so that you can get it in the 2.5 release (master is for 3.0, it won't be compatible with 2.5).

I will adapt this as soon as we agree on the next point.

> would be better in https://github.com/ezsystems/ezplatform-xmltext-fieldtype... after all, xmltext isn't meant to remain supported. It would also make it easier to distribute it to previous releases where graphql is compatible

I was considering this, of course. The tipping point however was `Field.types.yaml`. It is loaded in this bundle, hardcoded, with no obvious way of augmenting it in `ezplatform-xmltext-fieldtype` or other places. Am I missing something?

bdunogier:
> I was considering this, of course. The tipping point however was Field.types.yaml. It is loaded in this bundle, hardcoded, with no obvious way of augmenting it in ezplatform-xmltext-fieldtype or other places. Am I missing something?

It is actually easy. Just create the file in the bundle's `Resources/config/graphql` directory. The types defined there will automatically be registered. Note that it requires some extra config with the next overblog/graphql-bundle version, but that doesn't affect the 1.x branch of our package. It is done that way in https://github.com/ezsystems/ezplatform-query-fieldtype/tree/vendor/src/Symfony/Resources/config/graphql.

dfritschy:
Closed in favour of https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/111

PR #71, opened 2019-11-28
EZP-31108: Provided common PHP CS Fixer configuration for internal use

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-31108


bdunogier:
I'll apply it to master as well.

PR #76, opened 2020-02-04
Added support for Field values argsBuilder option

bdunogier:
> Required by https://github.com/ezsystems/ezplatform-query-fieldtype/pull/20

Add a `FieldDefinitionArgsMapper` interface that allows to define custom arguments for a field value. Mappers that implement it must be tagged with `ezplatform_graphql.field_definition_args_mapper`, with a `fieldtype` attribute set to the fieldtype's identifier.

For transparency, the interface is implemented by the `Configurable` and `Default` field definition mappers.

### 2.0
In 2.0, the interface is merged into the regular `FieldDefinitionMapper` interface, and the service tag is removed.

### DIFF
--- a/src/DependencyInjection/Compiler/FieldDefinitionArgsBuilderMappersPass.php
+++ b/src/DependencyInjection/Compiler/FieldDefinitionArgsBuilderMappersPass.php
@@ -1,0 +1,51 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformGraphQL\DependencyInjection\Compiler;
+
+use EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Worker\FieldDefinition\AddFieldValueToDomainContent;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Handles the ezplatform_graphql.field_definition_args_builder_mapper service tag.
+ * Injects tagged services as the $argsMapper argument of the AddFieldValueToDomainContent Field Definition Worker.
+ *
+ * @deprecated deprecated since 1.0, will be removed in 1.0. Use the FieldDefinitionArgsBuilderMapper interface in your decorating mapper.
+ */
+class FieldDefinitionArgsBuilderMappersPass implements CompilerPassInterface
+{
+    const TAG = 'ezplatform_graphql.field_definition_args_builder_mapper';
+    const ATTRIBUTE = 'fieldtype';
+
+    public function process(ContainerBuilder $container)
+    {
+        $mappers = [];
+        foreach ($container->findTaggedServiceIds(self::TAG) as $id => $tags) {
+            foreach ($tags as $tag) {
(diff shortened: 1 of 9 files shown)


src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldValueToDomainContent.php
```diff
@@ -37,10 +43,16 @@ public function work(Builder $schema, array $args)
...
     {
-        return [
+        $definition = [
             'type' => $this->fieldDefinitionMapper->mapToFieldValueType($fieldDefinition),
             'resolve' => $this->fieldDefinitionMapper->mapToFieldValueResolver($fieldDefinition),
         ];
+
+        if (isset($this->argsMappers[$fieldDefinition->fieldTypeIdentifier])) {
+            $definition['argsBuilder'] = $this->argsMappers[$fieldDefinition->fieldTypeIdentifier]->mapToFieldValueArgs($fieldDefinition);
+        }
```
adamwojs: If you like to merge `FieldDefinitionArgsMapper` interface into `FieldDefinitionMapper` in 2.0 maybe we could go with the following approach in 1.0: 

```php
        if ($this->fieldDefinitionMapper instanceof FieldDefinitionArgsMapper) {
            $definition['argsBuilder'] = $this->fieldDefinitionMapper->mapToFieldValueArgs($fieldDefinition);
        }
```

* `ezplatform_graphql.field_definition_args_mapper` tag and compiler pass will be not needed
* It will force users to implement `FieldDefinitionArgsMapper` together with `FieldDefinitionMapper`. When  `mapToFieldValueArgs ` will added to `FieldDefinitionMapper` in most cases will be enought to remove `implements FieldDefinitionArgsMapper` during the migration.

bdunogier:
Renamed the feature to "args builder mapper", as it is what it does.

src/Schema/Domain/Content/Worker/FieldDefinition/AddFieldValueToDomainContent.php
bdunogier: Note that there are third party packages implementing that feature already (matrix, page...), and we need to keep BC on it.

src/DependencyInjection/Compiler/FieldDefinitionArgsBuilderMappersPass.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Handles the ezplatform_graphql.field_definition_args_builder_mapper service tag.
+ * Injects tagged services as the $argsMapper argument of the AddFieldValueToDomainContent Field Definition Worker.
+ *
+ * @deprecated deprecated since 1.0, will be removed in 1.0. Use the FieldDefinitionArgsBuilderMapper interface in your decorating mapper.
+ */
+class FieldDefinitionArgsBuilderMappersPass implements CompilerPassInterface
+{
+    const TAG = 'ezplatform_graphql.field_definition_args_builder_mapper';
```
adamwojs:
```suggestion
    private const TAG = 'ezplatform_graphql.field_definition_args_builder_mapper';
```

src/DependencyInjection/Compiler/FieldDefinitionArgsBuilderMappersPass.php
```diff
@@ -0,0 +1,51 @@
...
+    const ATTRIBUTE = 'fieldtype';
```
adamwojs:
```suggestion
    private const ATTRIBUTE = 'fieldtype';
```

PR #82, opened 2020-04-23
EZP-31597: Handled missing Content Types when generating Schema

SerheyDolgushev:
More details at https://jira.ez.no/browse/EZP-31597

Also, it might worth showing errors like this during the schema update, but it would require injecting the logger into the mappers.

@bdunogier


alongosz:
Oh Guys... :joy: Please do not catch Exception from Core when an API one is available. 1st and 3rd party packages should never do that (1st party could throw core ATM though, because there's no other way).

I've fixed that and failing Travis for you when rebasing onto 2.0. Force-pushed.

alongosz:
Merged up to `master` via 6043218ac69ae1651baf1d7e3388839fae1a0a8b.

Thank you @SerheyDolgushev :tada:

PR #84, opened 2020-08-12
Fixing image variations resolver

philipp-heyse:
I had issues fetching multiple image variations in a single request 
(Array to String Conversion)

Upon debugging i figured the field type resolver is wrong


bdunogier:
Indeed, it looks legit, thank you.

I'm mostly using this syntax, and didn't really use the multiple one:
```
small_variation: variation(identifier: "small") { ... }
medium_variation: variation(identifier: "medium") { ... }
```

PR #87, opened 2020-10-07
Enhanced location handling

bdunogier:
- change the main item in use to `Location` instead of `Content`
- apply tree root settings

### TODO
- [x] Update collection resolver to return locations as well
- [x] In multi-location + multi-site cases, check that `_location` and `_url` refer to the relevant location
- [x] Handle root location prefix when loading by urlAlias (See [UrlAliasRouter](https://github.com/ezsystems/ezplatform-kernel/blob/25f4e4a991167a139210361bc9326d2fa67350dc/eZ/Bundle/EzPublishCoreBundle/Routing/UrlAliasRouter.php#L47-L60]))
- [x] Handle root location prefix when generating URL aliases


bdunogier:
Replaced by #90.

PR #89, opened 2020-11-02
EZP-32107: Impl. visitor for Score sort clause

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-32107

### Description

Added support for sorting by relevance score.


lserwatka:
Is this for 3.3 or 3.2?

adamwojs:
@lserwatka 3.3

mnocon (approved):
The SortClause can be used, but it has no effect - GraphQL performs a query with "filter" by default, which ignores Scoring: https://issues.ibexa.co/browse/EZP-32133

PR #90, opened 2020-11-25
Enhanced location handling (EZP-32182)

bdunogier:
A major refactoring of the GraphQL API that bumps it to v3.0.

`*Content` as well as the whole `DomainContent` terminology is replaced with a concept of `Item`, where an `Item` is the combination of a content and a location. Example: "ArticleContent" becomes "ArticleItem".

The goal is to add a concept of "current location" so that from any retrieved item, we can get a location (and its URL). It uses a Location Guesser API that given a Content item will guess a Location for it. It allows to keep the API content centric while allowing to work with locations in a smart way.

## Features
Illustration of the changes below:
- there are two public siteaccesses, `site1` and `site2`
- the query is made from `site1`.
- `folderBlog` 135 has two locations, one in `site1`, one in `site2`.
- `folderBlog.posts` is a query field that returns the blog posts that are children of the blog's current location.
- `blog.relatedBlogPosts` contains relations to both both the `site1` _and_ `site2` subtrees
Result:
- only the posts from the location in site1 are returned
- their URL is relative to `site1` (no `/site-1` prefix)
- the URL for the related blog post that is in `site2` is an absolute one, since it is on another siteaccess

```
{
  content {
    folderBlog(contentId: 135) {
      _name
      _url
      posts {
        edges {
          node {
            _url
            relatedBlogPosts {
              _url
            }
          }
        }
      }
    }
  }
}
```

```
{
  "data": {
    "content": {
      "folderBlog": {
        "_name": "Blog",
        "_url": "/blog",
        "posts": {
          "edges": [
            {
              "node": {
                "_url": "/blog/blog-post-site-2",
                "relatedBlogPosts": [
                  {
                    "_url": "https://site1.ezcommerce.local/blog/blog-post-site-1"
                  },
                  {
                    "_url": "/blog/friday-tech-beer"
                  }
                ]
              }
            },
            {
              "node": {
                "_url": "/blog/friday-tech-beer",
                "relatedBlogPosts": []
              }
            }
          ]
        }
      }
    }
  }
}
```
### Tree root support
Tree root settings are now taken into account automatically:
- only items that have a valid location in the current siteaccess are returned (e.g. within the subtree location id or within excluded paths)
- url aliases are stripped from the tree root prefix (e.g. "/site-1/item" => "/item")
- items can be searched by url alias without the tree root prefix ("/item" to search for "/site-1/item" when on siteaccess site1)

### Item objects
When working with content, and Item value object is used. It references either a location or a content, and guesses the other:
- If it is built with a Location, the content is retrieved from the Location.
- If it is built with a Content, the Location is _guessed_ using the Location Guesser API

A generic resource, `item`, with the same arguments than `article` or `folder`, is also available. It returns the item object, and its type can be tested in the query. It is meant to limit the usage of the `_repository { location` resource.

### Location guesser API
Given a Content item, the LocationGuesser will return the "best" location for it. It uses a set of filters on the content item's locations:
- only locations valid for the current siteaccess
- if several, the main one.

More of those filters could be added:
- within a given subtree (would require that a "Context" is provided)
- lowest priority
- etc.

### Siteaccess guesser API
A specific ItemFactory, `$relatedItemFactory`, will accept locations from other siteaccesses. When an URL for such an item is requested, it will be generated for the Item's siteaccess, using an absolute URL if necessary.

### Other minor changes
- The `id` argument for loading an item is replaced with `contentId`. `locationRemoteId` and `urlAlias` have also been added as arguments

### TODO
- [x] Fix CS
- [ ] Apply the same change to the mutation terminology (`createArticleContent` -> `createArticleItem` etc)
- [ ] Apply to type handling as well ?
- [ ] Change `Item { _locations }` so that it returns the locations that are valid for the item
- [ ] Add a shortcut to get the url alias for a location. Example: `Item { _locations { _url } }`
- [x] Add an `item` resource that returns an item independently of its type (use-case: search for any item with an URL alias)
- [x] Decide what to do with Related Items Locations: do we also restrict the results to the current siteaccess ? Or do we want some way to allow _external locations_ ? Is this a per-field thing ? Or should we just follow permissions (e.g. if you're allowed to view it, you can get a link to it) ?


src/GraphQL/ItemFactory.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        $this->locationGuesser = $locationGuesser;
+    }
+
+    public function fromContent(Content $content): Item
+    {
+        return new Item($this->locationGuesser, null, $content);
+    }
+
+    public function fromLocation(Location $location)
```
adamwojs, line 30:
```suggestion
    public function fromLocation(Location $location): Item
```

src/GraphQL/InputMapper/SearchQueryMapper.php
```diff
@@ -6,24 +6,53 @@
...
-    public function mapInputToQuery(array $inputArray)
+    public function mapInputToQuery(array $inputArray): Query
     {
         $query = new Query();
+        $this->mapInput($query, $inputArray);
+
+        return $query;
+    }
+
+    private function mapInput($query, array $inputArray)
```
adamwojs, line 47:
```suggestion
    private function mapInput(Query $query, array $inputArray): void
```

src/GraphQL/Resolver/ItemResolver.php
```diff
@@ -0,0 +1,237 @@
...
+            ['filter' => new Query\Criterion\ContentId($destinationContentIds)]
+        ));
+
+        if ($multiple) {
+            return array_map(
+                function ($contentId) use ($contentItems) {
+                    return $contentItems[array_search($contentId, array_column($contentItems, 'id'))];
+                },
+                $destinationContentIds
+            );
```
adamwojs, line 162: At the moment whole expression could be reduced to `return $contentItems` Did you mean 

```
return array_map(
    function (Content $content): Item {
        return $this->itemFactory->fromContent($content);
    },
    $contentItems
);     
```

?

src/GraphQL/Resolver/ItemResolver.php
```diff
@@ -0,0 +1,237 @@
...
+
+        return $paginator->auto(
+            $args,
+            function () use ($query) {
+                return $this->contentLoader->count($query);
+            }
+        );
+    }
+
+    public function resolveItemType(Location $location)
```
adamwojs, line 196:
```suggestion
    public function resolveItemType(Location $location): string
```

src/GraphQL/Resolver/ItemResolver.php
```diff
@@ -0,0 +1,237 @@
...
+        $typeName = $this->makeDomainContentTypeName(
+            $this->contentTypeLoader->load($location->contentInfo->contentTypeId)
+        );
+
+        return  ($this->typeResolver->hasSolution($typeName))
+            ? $typeName
+            : 'UntypedContent';
+    }
+
+    private function makeDomainContentTypeName(ContentType $contentType)
```
adamwojs, line 207:
```suggestion
    private function makeDomainContentTypeName(ContentType $contentType): string
```

src/GraphQL/Resolver/ItemResolver.php
```diff
@@ -0,0 +1,237 @@
...
+        return $converter->denormalize($contentType->identifier) . 'Content';
+    }
+
+    /**
+     * @return \eZ\Publish\API\Repository\LocationService
+     */
+    private function getLocationService()
+    {
+        return $this->repository->getLocationService();
+    }
```
adamwojs, line 220:
```suggestion
    private function getLocationService(): LocationService
    {
        return $this->repository->getLocationService();
    }
```

src/Exception/MultipleValidLocationsException.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param \eZ\Publish\API\Repository\Values\Content\Location[] $locations
+     */
+    public function __construct(Content $content, array $locations)
+    {
+        $this->locations = $locations;
+        parent::__construct(
+            sprintf(
+                'Could not determine which location to return for content with id %s. Possible candidates: %s)',
+                $content->id,
+                implode(',', array_map(function (Location $location) { return $location->pathString; }, $locations))
```
adamwojs, line 37: Check if we could simplify array\_map to:

```
implode(',', array_column($locations, 'pathString'))
```

src/Exception/MultipleValidLocationsException.php
```diff
@@ -0,0 +1,55 @@
...
+     */
+    private $content;
+
+    /**
+     * @param \eZ\Publish\API\Repository\Values\Content\Content $content
+     * @param \eZ\Publish\API\Repository\Values\Content\Location[] $locations
+     */
+    public function __construct(Content $content, array $locations)
+    {
+        $this->locations = $locations;
```
adamwojs, line 32: Can we merge properties assignments into one block (after parent constructor call)?

src/GraphQL/Resolver/ItemResolver.php
```diff
@@ -0,0 +1,237 @@
...
+        $contentType = $item->getContentInfo()->getContentType();
+
+        if ($contentType->identifier !== $contentTypeIdentifier) {
+            throw new UserError("Content {$item->getContentInfo()->id} is not of type '$contentTypeIdentifier'");
+        }
+
+        return $item;
+    }
+
+    public function resolveItemFieldValue(Item $item, $fieldDefinitionIdentifier)
```
adamwojs, line 138:
```suggestion
    public function resolveItemFieldValue(Item $item, $fieldDefinitionIdentifier): Field
```

src/GraphQL/Resolver/LocationGuesser/LocationGuess.php
```diff
@@ -0,0 +1,50 @@
...
+
+namespace EzSystems\EzPlatformGraphQL\GraphQL\Resolver\LocationGuesser;
+
+use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\Location;
+use EzSystems\EzPlatformGraphQL\Exception;
+
+class LocationGuess
+{
+    private $content;
```
adamwojs, line 17: Nitpick: missing `/** @var ...`

src/GraphQL/Resolver/LocationGuesser/LocationGuess.php
```diff
@@ -0,0 +1,50 @@
...
+
+    private $locations;
```
adamwojs, line 19: Nitpick: missing `/** @var ...`

src/GraphQL/Resolver/LocationGuesser/LocationList.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    private $locations;
+
+    public function __construct(Content $content)
+    {
+        $this->content = $content;
+        $this->locations = new SplObjectStorage();
+    }
+
+    public function addLocation(Location $location)
```
adamwojs, line 37:
```suggestion
    public function addLocation(Location $location): void
```

src/GraphQL/Resolver/LocationGuesser/LocationList.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location
+     */
+    public function hasOneLocation(): bool
+    {
+        return count($this->locations) === 1;
+    }
+
+    public function filter(Location $location)
```
adamwojs, line 72:
```suggestion
    public function removeLocation(Location $location): void
```

src/GraphQL/Resolver/LocationGuesser/TreeRootLocationFilter.php
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * @param array $path
+     * @param array $rootPath
+     *
+     * @return bool
+     */
+    private function containsRootPath(array $path, array $rootPath): bool
+    {
+        return array_slice($path, 0, count($rootPath)) === $rootPath;
+    }
```
adamwojs, line 122: We should have Path VO into PAPI to avoid re-implementing path operation at-hoc...

src/GraphQL/Resolver/LocationGuesser/TreeRootLocationFilter.php
```diff
@@ -0,0 +1,123 @@
...
+
+        return false;
+    }
+
+    /**
+     * @param array $path
+     * @param array $rootPath
+     *
+     * @return bool
+     */
```
adamwojs, line 118:
```suggestion
    /**
     * @param string[] $path
     * @param string[] $rootPath
     */
```

src/GraphQL/Resolver/UrlAliasResolver.php
```diff
@@ -51,4 +68,60 @@ public function resolveUrlAliasType(URLAlias $urlAlias)
                 return $this->typeResolver->resolve('VirtualUrlAlias');
         }
     }
+
+    public function resolveLocationUrlAlias(Location $location)
```
adamwojs, line 72:
```suggestion
    public function resolveLocationUrlAlias(Location $location): ?string
```

src/GraphQL/Value/Item.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /** @var \EzSystems\EzPlatformGraphQL\GraphQL\Resolver\LocationGuesser\LocationGuesser */
+    private $locationGuesser;
+
+    public function __construct(LocationGuesser $locationGuesser, ?Location $location = null, ?Content $content = null)
+    {
+        if ($location === null && $content === null) {
+            // @todo find a better exception
+            throw new \Exception('Constructing an item requires one of location or content');
+        }
```
adamwojs, line 33: ... or mark constructor as private/protected and force class instantiation using named constructors: 

```
public static fromContent(LocationGuesser $locationGuesser, Content $content): self;
```

```
public static fromLocation(LocationGuesser $locationGuesser, Content $location): self;
```

src/GraphQL/Resolver/ItemResolver.php
bdunogier: If I recall correctly (I wrote that a while ago) the goal was to return the items in the order they are defined in the relation field. And if I recall correctly (again), the search result doesn't respect the order the id are passed in.

src/GraphQL/Resolver/LocationGuesser/TreeRootLocationFilter.php
bdunogier: Fine by me.

bdunogier: We could even go further, and have an official API for getting the URL alias for a location for a given siteaccess.

src/Exception/MultipleValidLocationsException.php
bdunogier: According to the docs, we can.

src/GraphQL/Resolver/ItemResolver.php
```diff
@@ -0,0 +1,237 @@
...
+        } elseif (isset($args['locationId'])) {
+            $item = $this->itemFactory->fromLocation(
+                $this->locationLoader->findById($args['locationId'])
+            );
+        } elseif (isset($args['locationRemoteId'])) {
+            $item = $this->itemFactory->fromLocation(
+                $this->locationLoader->findByRemoteId($args['locationRemoteId'])
+            );
+        } else {
+            throw new UserError('Missing required argument id, remoteId or locationId');
```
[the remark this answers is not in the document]
bdunogier: Indeed. And it looks like I have lost my `urlAlias` argument too.

bdunogier:
We should be green. There are still todo items above, but they may also be taken care of in a follow-up. I'll work on them anyway, but if we could consider this PR for merging, it would be nice.

bdunogier:
To reviewers, I'd like to merge this as soon as possible and iterate further. There are remaining tasks, but it's a large refactoring, and doing it all in one pull-requests will make it difficult.

src/GraphQL/Resolver/ItemResolver.php
adamwojs: > the search result doesn't respect the order the id are passed in.

True, especially in multi node setup.

bdunogier:
Summary of the plan for this:
- this is a major change, a BC break from GraphQL 2.x
- the idea is/was to release a 3.0-beta that *can* be installed on the LTS, and will be stabilized for the 4.0 release

If the plan doesn't work, please let me know. But releasing a BC break of that package in the LTS doesn't sound correct.

PR #98, opened 2021-03-22
EZP-32261: Added field name override

mateuszbieniek:
This PR adds an option to define field name overrides to avoid issues like:
https://issues.ibexa.co/browse/EZP-32261

### DIFF
--- a/src/Resources/config/default_settings.yaml
+++ b/src/Resources/config/default_settings.yaml
@@ -1,4 +1,6 @@
 parameters:
+  ezplatform_graphql.schema.content.field_name.override:
+    id: id_
   ezplatform_graphql.schema.content.mapping.field_definition_type:
     ezauthor:
       value_type: "[AuthorFieldValue]"
--- a/src/Schema/Domain/Content/NameHelper.php
+++ b/src/Schema/Domain/Content/NameHelper.php
@@ -18,9 +18,15 @@ class NameHelper
      */
     private $caseConverter;
 
-    public function __construct()
+    /**
+     * @var array
+     */
+    private $fieldNameOverrides ;
+
+    public function __construct(array $fieldNameOverrides)
     {
         $this->caseConverter = new CamelCaseToSnakeCaseNameConverter(null, false);
+        $this->fieldNameOverrides = $fieldNameOverrides;
     }
 
     public function domainContentCollectionField(ContentType $contentType)
@@ -90,7 +96,14 @@ class NameHelper
 
     public function fieldDefinitionField(FieldDefinition $fieldDefinition)
     {
-        return lcfirst($this->toCamelCase($fieldDefinition->identifier));
+        $fieldName = lcfirst($this->toCamelCase($fieldDefinition->identifier));
+
+        // Workaround for https://issues.ibexa.co/browse/EZP-32261
+        if (array_key_exists($fieldName, $this->fieldNameOverrides)) {
+            return $this->fieldNameOverrides[$fieldName];
+        }
+
+        return $fieldName;
     }
 
     private function toCamelCase($string)
--- a/src/Schema/Domain/Content/NameHelper.php
+++ b/src/Schema/Domain/Content/NameHelper.php
@@ -10,6 +10,7 @@ use eZ\Publish\API\Repository\Values\ContentType\ContentType;
 use eZ\Publish\API\Repository\Values\ContentType\ContentTypeGroup;
 use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
 use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter;
 
 class NameHelper
@@ -29,10 +30,10 @@ class NameHelper
      */
     private $fieldNameOverrides;
 
-    public function __construct(LoggerInterface $logger, array $fieldNameOverrides)
+    public function __construct(array $fieldNameOverrides, LoggerInterface $logger = null)
     {
         $this->caseConverter = new CamelCaseToSnakeCaseNameConverter(null, false);
-        $this->logger = $logger;
+        $this->logger = $logger ?? new NullLogger();
         $this->fieldNameOverrides = $fieldNameOverrides;
     }
 
--- a/src/Schema/Domain/Content/NameHelper.php
+++ b/src/Schema/Domain/Content/NameHelper.php
@@ -21,7 +21,7 @@ class NameHelper
     /**
      * @var array
      */
-    private $fieldNameOverrides ;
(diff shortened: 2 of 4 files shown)


src/Schema/Domain/Content/NameHelper.php
```diff
@@ -18,9 +18,15 @@ class NameHelper
      */
     private $caseConverter;
 
-    public function __construct()
+    /**
+     * @var array
```
adamwojs, line 22:
```suggestion
     * @var string[]
```

src/Resources/config/default_settings.yaml
```diff
@@ -1,4 +1,6 @@
 parameters:
+  ezplatform_graphql.schema.content.field_name.override:
+    id: id_
```
adamwojs, line 3: Is `id` the only one field which might generate name collision?

src/Schema/Domain/Content/NameHelper.php
```diff
@@ -90,7 +96,14 @@ public function domainGroupField(ContentTypeGroup $contentTypeGroup)
 
     public function fieldDefinitionField(FieldDefinition $fieldDefinition)
     {
-        return lcfirst($this->toCamelCase($fieldDefinition->identifier));
+        $fieldName = lcfirst($this->toCamelCase($fieldDefinition->identifier));
+
+        // Workaround for https://issues.ibexa.co/browse/EZP-32261
+        if (\array_key_exists($fieldName, $this->fieldNameOverrides)) {
+            return $this->fieldNameOverrides[$fieldName];
```
adamwojs, line 103: IMHO is worth to warn user at this point that name collision happened.

src/Resources/config/default_settings.yaml
mateuszbieniek: That's the only field that I'm aware of.

adamwojs: OK, but please rememeber ot document this behaviour. // cc @ezsystems/documentation-team

Steveb-p (approved):
Ideally we would rather prevent declaring fields as "id", but I guess that's the next best thing.

I would suggest making it forbidden to do so in 4.x, treating "id" (along with uppercase variations) like a PHP keyword - just not allow it. WDYT?

mateuszbieniek:
> Ideally we would rather prevent declaring fields as "id", but I guess that's the next best thing.
> 
> I would suggest making it forbidden to do so in 4.x, treating "id" (along with uppercase variations) like a PHP keyword - just not allow it. WDYT?

I agree, but we still have to keep in mind, that we have to prepare some migration scenario for the customers upgrading from (for example) 2.5 LTS, where `id` was totally valid identifier for the fields.

And of course some entries in the documentation.

src/Schema/Domain/Content/NameHelper.php
mateuszbieniek: Added a warning level log entry so it will be shown in the output of
```
php bin/console ezplatform:graphql:generate-schema
```

src/Schema/Domain/Content/NameHelper.php
```diff
@@ -18,9 +19,21 @@ class NameHelper
      */
     private $caseConverter;
 
-    public function __construct()
+    /**
+     * @var \Psr\Log\LoggerInterface
+     */
+    private $logger;
```
Steveb-p, line 25: Use `LoggerAwareTrait` and `LoggerAwareInterface` instead.

src/Schema/Domain/Content/NameHelper.php
```diff
@@ -18,9 +19,21 @@ class NameHelper
...
+
+    /**
+     * @var string[]
+     */
+    private $fieldNameOverrides;
+
+    public function __construct(LoggerInterface $logger, array $fieldNameOverrides)
     {
         $this->caseConverter = new CamelCaseToSnakeCaseNameConverter(null, false);
+        $this->logger = $logger;
```
Steveb-p, line 35: Allow logger to be null and initialize it to `NullLogger` if done so. This will make the eventual testing easier.

```suggestion
    public function __construct(array $fieldNameOverrides, ?LoggerInterface $logger = null)
    {
        $this->caseConverter = new CamelCaseToSnakeCaseNameConverter(null, false);
        $this->logger = $logger ?? new NullLogger();
```
Don't forget to import the class if this suggestion is acceptable for you.

Steveb-p (changes requested):

Steveb-p (approved):

adamwojs (approved):

src/Schema/Domain/Content/NameHelper.php
```diff
@@ -109,7 +138,7 @@ private function pluralize($name)
         }
 
         if (substr($name, -1) === 'y') {
-            if (in_array(substr($name, -2, 1), ['a', 'e', 'i', 'o', 'u'])) {
+            if (\in_array(substr($name, -2, 1), ['a', 'e', 'i', 'o', 'u'])) {
```
alongosz, line 141: CS: you should rather import it with `use function in_array;`

src/Schema/Domain/Content/NameHelper.php
```diff
@@ -124,7 +153,7 @@ private function pluralize($name)
             return substr($name, 0, -2) . 'i';
         }
 
-        if (in_array(substr($name, -2), ['on', 'um'])) {
+        if (\in_array(substr($name, -2), ['on', 'um'])) {
```
alongosz, line 156: CS: the same remark as above

alongosz (approved):

src/Schema/Domain/Content/NameHelper.php
mateuszbieniek: Running `php-cs-fixer fix` drops `use function in_array;` and adds `\in_array` :man_shrugging:

tomaszszopinski (approved):
QA approved on Ibexa DXP 3.2.5-dev & v3.2.5 with patch & diff.

lserwatka:
Could you merge it up?

mateuszbieniek:
Merged to 3.0 in https://github.com/ezsystems/ezplatform-graphql/commit/7d75c686566cf91239799b4e7888f7288e701eaa

mateuszbieniek:
Merged to master in https://github.com/ezsystems/ezplatform-graphql/commit/86f58702161da3060333459c55a677bf95d16d5f

src/Resources/config/default_settings.yaml
bdunogier: There are actually more, but they are all prefixed with `_`, making collisions less likely: https://github.com/ezsystems/ezplatform-graphql/blob/86f58702161da3060333459c55a677bf95d16d5f/src/Resources/config/graphql/DomainContent.types.yaml#L8-L31.

PR #101, opened 2021-05-14
IBX-354: Used thumbnail strategy for thumbnail generation

ViniTou:
https://issues.ibexa.co/browse/IBX-354

frontend part adjustments: 
https://github.com/ezsystems/ezplatform-admin-ui/pull/1756

### DIFF
--- a/src/GraphQL/Resolver/ContentThumbnailResolver.php
+++ b/src/GraphQL/Resolver/ContentThumbnailResolver.php
@@ -40,6 +40,7 @@ final class ContentThumbnailResolver
             'width' => $thumbnail->width,
             'height' => $thumbnail->height,
             'mimeType' => $thumbnail->mimeType,
+            'alternativeText' => '',
         ];
     }
 }
--- a/src/GraphQL/Resolver/ContentThumbnailResolver.php
+++ b/src/GraphQL/Resolver/ContentThumbnailResolver.php
@@ -6,45 +6,18 @@
  */
 namespace EzSystems\EzPlatformGraphQL\GraphQL\Resolver;
 
-use Exception;
 use eZ\Publish\API\Repository\Values\Content\Content;
-use eZ\Publish\API\Repository\Values\Content\Field;
-use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
-use eZ\Publish\Core\FieldType;
-use eZ\Publish\SPI\Variation\VariationHandler;
-use EzSystems\EzPlatformGraphQL\GraphQL\DataLoader\ContentLoader;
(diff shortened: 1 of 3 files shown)


src/GraphQL/Resolver/ContentThumbnailResolver.php
```diff
@@ -6,88 +6,40 @@
...
-
         return [
-            'uri' => $thumbnailVariation->uri,
-            'width' => $thumbnailVariation->width,
-            'height' => $thumbnailVariation->height,
-            'alternativeText' => $imageField->value->alternativeText,
+            'uri' => $thumbnail->resource,
+            'width' => $thumbnail->width,
+            'height' => $thumbnail->height,
+            'mimeType' => $thumbnail->mimeType,
```
Steveb-p, line 42: Missing `alternativeText` in this array might cause issues. Is this intentional?

ViniTou: >intentional
yes. Our thumbnail object (\eZ\Publish\API\Repository\Values\Content\Thumbnail) does not have alternative text atm, it would have to be implemented for every `\eZ\Publish\SPI\Repository\Strategy\ContentThumbnail\Field\ThumbnailStrategy` we have.

On the other hand, you maybe right and till it will happen maybe we should just put empty string there.

alongosz (approved):

src/GraphQL/Resolver/ContentThumbnailResolver.php
```diff
@@ -6,88 +6,41 @@
...
-        try {
-            $imageField = $this->getThumbnailImageField($content);
-        } catch (Exception $e) {
-            return null;
-        }
+        $thumbnail = $this->thumbnailStrategy->getThumbnail(
+            $content->getContentType(),
+            $content->getFields(),
+            $content->getVersionInfo()
+        );
```
adamwojs, line 32: hmmm I've got feeling that `try ... catch` statement should stay here (with proper error logging)

adamwojs (approved):

src/GraphQL/Resolver/ContentThumbnailResolver.php
ViniTou: From what I see we are not catchin it inside `\eZ\Publish\Core\Repository\Mapper\ContentDomainMapper::buildContentDomainObject` and we did not had problems with it till now as exceptions are catched (without login tho) inside specific strategies.

adamwojs: OK

Steveb-p (approved):

ViniTou:
merged up

PR #102, opened 2021-06-24
variations → variation; alias → identifier; html → html5 

adriendupuis:
- For only one variation, using `variation` instead of `variations` could be better.
- `variation` and `variations` both use `identifier` argument, not `alias`.
- RichText has `html5` instead of `html`

### DIFF
--- a/doc/domain_schema.md
+++ b/doc/domain_schema.md
@@ -30,7 +30,7 @@ Queries look like this:
       body { html }
       image {
         name
-        variations(alias: large) { uri }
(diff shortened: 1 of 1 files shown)


doc/domain_schema.md
```diff
@@ -27,10 +27,10 @@ Queries look like this:
   content {
     articles {
       title
-      body { html }
+      body { html5 }
```
adriendupuis, line 30:
```suggestion
      body { html: html5 }
```

PR #116, opened 2021-12-03
IBX-1565: Refactor schema's types' names validation

barw4:
| **JIRA issue**                          | [IBX-1565](https://issues.ibexa.co/browse/IBX-1565)
| **Type**                                   | bug
| **Target version** | `v2.5`

`Content Type Groups` don't have an identifier that matches the GraphQL standard, they have only names. Therefore the problem arises in every Group which has some special chars and the schema is not generated at all for them, even though the name would be transformed.

The validation should occur after the identifiers have been transformed according to the standard, not before. This solution also simplifies code and avoids repeats.

### DIFF
--- a/src/Schema/Builder/SchemaBuilder.php
+++ b/src/Schema/Builder/SchemaBuilder.php
@@ -7,10 +7,25 @@
 namespace EzSystems\EzPlatformGraphQL\Schema\Builder;
 
 use EzSystems\EzPlatformGraphQL\Schema\Builder as SchemaBuilderInterface;
+use Ibexa\GraphQL\Schema\Domain\NameValidator;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
+use Psr\Log\NullLogger;
 
-class SchemaBuilder implements SchemaBuilderInterface
+class SchemaBuilder implements SchemaBuilderInterface, LoggerAwareInterface
 {
+    use LoggerAwareTrait;
+
     private $schema = [];
+
+    /** @var \Ibexa\GraphQL\Schema\Domain\NameValidator */
+    private $nameValidator;
+
+    public function __construct(NameValidator $nameValidator)
+    {
+        $this->nameValidator = $nameValidator;
+        $this->logger = new NullLogger();
+    }
 
     public function getSchema(): array
     {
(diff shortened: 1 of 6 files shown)


src/Schema/Builder/SchemaBuilder.php
```diff
@@ -150,4 +177,13 @@ public function hasEnum($enum): bool
     {
         return $this->hasType($enum);
     }
+
+    private function generateInvalidGraphQLNameWarning(string $type, string $name): void
+    {
+        $message = "Skipping schema generation for %s with identifier '%s'. "
+            . 'Please rename given %s according to GraphQL specification '
+            . '(http://spec.graphql.org/June2018/#sec-Names)';
```
konradoboza: Not sure about maintaining the URL within this message - we already have a newer specification release: https://spec.graphql.org/October2021/#sec-Names. It doesn't seem to be tied to any GraphQL package used by DXP, so maybe we can tweak it a little, e.g.:
```suggestion
            . '(http://spec.graphql.org/[latest-release]/#sec-Names)';
```
I see it was added before your change and before the newest release of the specification (https://github.com/ezsystems/ezplatform-graphql/pull/109) but I will personally make it a little tidier anyway. @adamwojs would you agree?

barw4: Tbh this message as you have noticed was just really copied and ideally, it should never be thrown because the check is done **after** the item renaming. We cannot expect the user to rename this item as this is generated later on when the whole schema is built.

Honestly, I would leave only `Skipping schema generation for %s with identifier '%s'.` as the user has no power at this point to update anything.

konradoboza: Fine by me. Nonetheless, I would give some more explanation though on why skipping took place. Maybe:
```
Skipping schema generation for %s with identifier '%s' as it stands against GraphQL specification. For more details see http://spec.graphql.org/[latest-release]/#sec-Names.
```

barw4: Done in https://github.com/ezsystems/ezplatform-graphql/pull/116/commits/01261bac86888d8b447b9d9e41813213ac1b29a6

bdunogier (approved):
Better than the previous approach, and more flexible since we can transform some other specific names that would right now be invalid.

konradoboza (approved):

adamwojs (approved):

tomaszszopinski (approved):
QA approved on eZPlatform 2.5.

barw4:
Merge into:
`2.3`: https://github.com/ezsystems/ezplatform-graphql/commit/a56b2f4baa638f8099f0124c6a10181ada98c1ea
`ibexa/graphql/main`: https://github.com/ibexa/graphql/commit/54d14ec09b531720183e4b814f2ad9926e55b656

PR #128, opened 2022-07-04
IBX-3095: Fixed BinaryFile fieldtype resolving

barw4:
| **JIRA issue**                          | [IBX-3095](https://issues.ibexa.co/browse/IBX-3095)
| **Type**                                   | bug
| **Target Ibexa version** | `v3.3`

### DIFF
--- a/src/Resources/config/graphql/Field.types.yaml
+++ b/src/Resources/config/graphql/Field.types.yaml
@@ -88,13 +88,10 @@ AuthorFieldValue:
             email:
                 type: "String"
 
-BinaryBaseFieldValue:
+BinaryFileFieldValue:
     type: object
     config:
         fields:
-            id:
-                type: "String"
-                resolve: "@=value.id"
             fileName:
                 type: "String"
                 resolve: "@=value.fileName"
@@ -107,14 +104,6 @@ BinaryBaseFieldValue:
             uri:
                 type: "String"
                 resolve: "@=value.uri"
-
-BinaryFileFieldValue:
-    type: object
-    config:
-        fields:
-            downloadCount:
-                type: Int
-    inherits: [BinaryBaseFieldValue]
 
 CountryFieldValue:
     type: object
@@ -210,7 +199,7 @@ MediaFieldValue:
                 type: Int
                 description: "Width of the media."
                 resolve: "@=value.width"
-    inherits: [BinaryBaseFieldValue]
+    inherits: [BinaryFileFieldValue]
 
 PriceFieldValue:
     type: object
--- a/src/Resources/config/graphql/Field.types.yaml
+++ b/src/Resources/config/graphql/Field.types.yaml
@@ -93,17 +93,20 @@ BinaryBaseFieldValue:
     config:
         fields:
             id:
-                type: String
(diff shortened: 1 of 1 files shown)


konradoboza (approved):

adamwojs (approved):

glye:
Please don't merge yet, pending comment from @mnocon about an interesting point, coming soon.

mnocon (commented):
The solution looks good, for the query:
```
{
  content {
    articles {
      edges {
        node {
          _name
          newEzbinaryfile8 {
            id
            fileName
            fileSize
            mimeType
            downloadCount
            uri
            mimeType
          }
        }
      }
    }
  }
}
```

the response looks like this:
```
{
  "data": {
    "content": {
      "articles": {
        "edges": [
          {
            "node": {
              "_name": "Test123123",
              "newEzbinaryfile8": {
                "id": "image/b44ee0a57fea81643701e0ab82bf1ddf.jfif",
                "fileName": "Crew Event 2019 original picture.jfif",
                "fileSize": 316062,
                "mimeType": "image/jpeg",
                "downloadCount": null,
                "uri": "/content/download/56/264?version=4"
              }
            }
          }
        ]
      }
    }
  }
}
```

The only concern I have is the `id` property (`"id":image/b44ee0a57fea81643701e0ab82bf1ddf.jfif",`). 

`b44ee0a57fea81643701e0ab82bf1ddf` is the filename on the disk (the file is stored in `public/var/site/storage/original/image/b44ee0a57fea81643701e0ab82bf1ddf`)

I know that in order to get the GraphQL response I need to have permissions to view that Content Item, so I can already download it - and most likely there's no real security risk here.

But the filename on the disk is for me an implementation detail that's not worth having in the API, and we had some security tickets related to filenames of uploaded files (https://issues.ibexa.co/browse/IBX-1656, https://issues.ibexa.co/browse/EZP-31139) so I've decided that it's best to double check whether we really want to include it in the response.

@barw4 @adamwojs @konradoboza @glye if you're ok with including the `id` property then I'm fine with as well, just want to make sure.

Other: the downloadCount property is still `null` - not sure if we should do something about it (and if it's a good idea to expose it)

konradoboza:
I agree with @mnocon, `id` property doesn't bring much value and we shouldn't expose it. The same stands for `downloadCount`.

glye:
Thanks @mnocon. All: Is there any usecase at all for including the `id`? If not, it's just noise. For instance, in our default vhost we do not grant read access to the `original` dir, so direct access from there is not supposed to be possible.

I agree about `downloadCount` as it provides info not just about the info but the system around it. It is more of an admin related property than a generically useful read property.

barw4:
`id` and `downloadCount` have been removed in https://github.com/ezsystems/ezplatform-graphql/pull/128/commits/7fb09f749d9785e9008fe79c8fa03afb60e33b92, also `BinaryBaseFieldValue` was removed as it was useless.

konradoboza (approved):

glye (approved):

mnocon (approved):
Looks good now, thanks!

PR #130, opened 2022-12-21
IBX-4542: Implemented `ImageAdditionalData` type

barw4:
| **JIRA issue**                          | [IBX-4542](https://issues.ibexa.co/browse/IBX-4542)
| **Type**                                   | improvement
| **Target Ibexa version** | `v3.3`

Example query:
```
{
  media {
    image (id: 261) {
      id,
      name,
      image {
        id,
        fileName,
        fileSize,
        additionalData {
          focalPointX,
          focalPointY
        }
      }
    }
  }
}
```

### DIFF
--- a/src/Resources/config/graphql/Field.types.yaml
+++ b/src/Resources/config/graphql/Field.types.yaml
@@ -36,6 +36,9 @@ ImageFieldValue:
             height:
                 type: "Int"
                 resolve: "@=value.height"
+            additionalData:
+                type: "ImageAdditionalData"
+                resolve: "@=value.additionalData"
             variations:
                 type: "[ImageVariation]"
                 args:
@@ -50,6 +53,17 @@ ImageFieldValue:
                         type: ImageVariationIdentifier!
                         description: "A variation identifier."
                 resolve: "@=resolver('ImageVariation', [value.value, args])"
+
+ImageAdditionalData:
+    type: object
+    config:
+        fields:
+            focalPointX:
+                type: "Int"
+                description: "The X value of focal point"
+            focalPointY:
+                type: "Int"
+                description: "The Y value of focal point"
 
 ImageVariation:
     type: object


konradoboza (approved):

bdunogier:
What about calling it `focalPoint` ?

src/Resources/config/graphql/Field.types.yaml
```diff
@@ -36,6 +36,9 @@ ImageFieldValue:
             height:
                 type: "Int"
                 resolve: "@=value.height"
+            additionalData:
```
bdunogier, line 39: What about calling it `focalPoint` ?

```
image {
  focalPoint {
    x
    y
  }
}
```

bdunogier (changes requested):

src/Resources/config/graphql/Field.types.yaml
barw4: @bdunogier shouldn't it be on par with what we have in the Image fieldtype? Those values are in fact stored in `additionalData['focalPointX']` and `additionalData['focalPointY']` so this is 1:1 representation.

adamwojs: Are focalPointX and focalPointY are available in ibexa/oss ?

barw4: @adamwojs they are not, however, querying them will just return `null`s like shown below:
image
This is the exact same behavior as if we wouldn't have a focal point set on `ibexa/content`.

adamwojs (approved):

micszo (approved):
Worked fine for me. 👍🏻 

Tested on Ibexa Commerce 3.3.30-dev.

src/Resources/config/graphql/Field.types.yaml
bdunogier: OK, fine :)

PR #137, opened 2024-02-05
IBX-7603: Added missing subfield for ezurl

mateuszdebinski:
Jira: https://issues.ibexa.co/browse/IBX-7603

If the ibexa.graphql.schema.should.extend.ezurl parameter is false (default), the graphQL query will look like before, i.e.:
```
{
  content {
    urlCT(contentId:68) {
      urlField
    }
  }
}
```

but if we set this parameter to true, the graphQL query will look like this:
```
{
  content {
    urlCT(contentId:68) {
      urlField {
        link,
        text
      }
    }
  }
}
```

### DIFF
--- a/src/Resources/config/services/schema.yaml
+++ b/src/Resources/config/services/schema.yaml
@@ -49,6 +49,12 @@ services:
         arguments:
             $innerMapper: '@EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\SelectionFieldDefinitionMapper.inner'
 
+    Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper:
+            decorates: EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\FieldDefinitionMapper
+            arguments:
+                $innerMapper: '@Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper.inner'
+                $shouldExtendUrlInputType: '%ibexa.graphql.schema.should.extend.ezurl%'
+
     EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Worker\ContentType\AddContentOfTypeConnectionToDomainGroup: ~
 
     EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Worker\ContentType\AddContentTypeToContentTypeIdentifierList: ~
--- a/src/Resources/config/services/schema.yaml
+++ b/src/Resources/config/services/schema.yaml
@@ -50,10 +50,10 @@ services:
             $innerMapper: '@EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\SelectionFieldDefinitionMapper.inner'
 
     Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper:
-            decorates: EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\FieldDefinitionMapper
-            arguments:
-                $innerMapper: '@Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper.inner'
-                $shouldExtendUrlInputType: '%ibexa.graphql.schema.should.extend.ezurl%'
+        decorates: EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\FieldDefinitionMapper
+        arguments:
+            $innerMapper: '@Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper.inner'
+            $shouldExtendUrlInputType: '%ibexa.graphql.schema.should.extend.ezurl%'
 
     EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Worker\ContentType\AddContentOfTypeConnectionToDomainGroup: ~
 
--- a/src/Schema/Domain/Content/Mapper/FieldDefinition/UrlFieldDefinitionMapper.php
+++ b/src/Schema/Domain/Content/Mapper/FieldDefinition/UrlFieldDefinitionMapper.php
@@ -1,16 +1,18 @@
 <?php
 
 /**
- * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
  * @license For full copyright and license information view LICENSE file distributed with this source code.
  */
+declare(strict_types=1);
+
 namespace Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition;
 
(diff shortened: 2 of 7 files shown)


konradoboza:
@mateuszdebinski could you please provide example query that will no longer work? The description is close to non-existent, it's hard to tackle the exact problem judging only by the schema adjustment.

mateuszdebinski:
> @mateuszdebinski could you please provide example query that will no longer work? The description is close to non-existent, it's hard to tackle the exact problem judging only by the schema adjustment.

@konradoboza  I added what it looks like now and what it will look like after the change, The current query will not work without specifying what fields you want to extract from the query

konradoboza (approved):
Thank you for providing examples @mateuszdebinski, now the topic is clear. While it's technically a BC break, I see that we are lacking `UrlFieldValue` anyway since that's how we handle fieldtypes in general when it comes to GraphQL schema. From my perspective it's an acceptable solution, but want to poke @bdunogier and maybe @alongosz for their opinions. An alternative that would spare us BC concerns would be creating another  dedicated GraphQL endpoint though.

bdunogier:
> creating another dedicated GraphQL endpoint though.
We can't really do that, the graphql endpoint is kind of "unique".

But BC is a problem here. If a project uses an `ezurl` field over GraphQL, their existing query _will_ break, period. I don't think it's acceptable.

GraphQL doesn't have anything for changing a primitive type to a custom one, afaik. The only way I can think of is some kind of feature flag. We add a custom mapper for that field type, and in the mapper, we either `String` or `UrlFieldValue` depending on the flag.

As it doesn't seem very likely that this is widely used, and in order to maximize the adoption of the new schema, we could disable the BC flag by default, _and document that it must be enabled if such a field type is used_. Of course, updating the client code is a better option.

bdunogier (changes requested):
BC concern, see https://github.com/ezsystems/ezplatform-graphql/pull/137#issuecomment-1983214138.

alongosz:
@bdunogier I don't know much about GraphQL. Is it possible to keep here BC? Meaning when old query is passed, returning old result? We need to have a solution here.

bdunogier:
Unfortunately, no, @alongosz, no BC features for changing a primitive type to an object. Hence my suggestion above for a feature flag.

bdunogier:
Let's complete this one, it's not gonna fix itself :)

The way to fix it is correct, but the BC break isn't. We can change the schema that way in v5.0, but not in 4.6. It is likely that this ezurl is used with GraphQL by existing apps.

We must add a feature flag / configuration directive that enables the new format, and ship it as disabled on 4.x. On 5.0, we can either ship the feature flag and enable it by default, or just change the schema. In any case, we should add a `@deprecated` flag to the current version of the field.

Is that okay for everybody ?

mateuszdebinski:
As we do not recommend using a lower PHP version than 8.1 and ultimately 7.4, I changed the dependency in composer for PHP to at least 7.4, I don't know if I can do that, so please let me know whether to restore support for 7.3

adamwojs:
Sorry, but we need to keep compatibility with PHP 7.3. Bumping minimum required version of package is a BC break.

bdunogier (approved):

src/Resources/config/services/schema.yaml
```diff
@@ -49,6 +49,12 @@ services:
         arguments:
             $innerMapper: '@EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\SelectionFieldDefinitionMapper.inner'
 
+    Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper:
+            decorates: EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\FieldDefinitionMapper
+            arguments:
+                $innerMapper: '@Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper.inner'
+                $shouldExtendUrlInputType: '%ibexa.graphql.schema.should.extend.ezurl%'
```
konradoboza:
```suggestion
    Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper:
        decorates: EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\FieldDefinitionMapper
        arguments:
            $innerMapper: '@Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition\UrlFieldDefinitionMapper.inner'
            $shouldExtendUrlInputType: '%ibexa.graphql.schema.should.extend.ezurl%'
```

src/Schema/Domain/Content/Mapper/FieldDefinition/UrlFieldDefinitionMapper.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
```
konradoboza, line 6:
```suggestion
 */
declare(strict_types=1);
```

src/Schema/Domain/Content/Mapper/FieldDefinition/UrlFieldDefinitionMapper.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Ibexa\GraphQL\Schema\Domain\Content\Mapper\FieldDefinition;
+
+use EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\DecoratingFieldDefinitionMapper;
+use EzSystems\EzPlatformGraphQL\Schema\Domain\Content\Mapper\FieldDefinition\FieldDefinitionMapper;
+use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
+
+class UrlFieldDefinitionMapper extends DecoratingFieldDefinitionMapper implements FieldDefinitionMapper
```
konradoboza:
```suggestion
final class UrlFieldDefinitionMapper extends DecoratingFieldDefinitionMapper implements FieldDefinitionMapper
```

src/Schema/Domain/Content/Mapper/FieldDefinition/UrlFieldDefinitionMapper.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    private const FIELD_TYPE_IDENTIFIER = 'ezurl';
+
+    /** @var bool */
+    private $shouldExtendUrlInputType;
```
konradoboza, line 20:
```suggestion
    private bool $shouldExtendUrlInputType;
```

mateuszdebinski: we can't do this because PHP is set minimum 7.3
"php": "^7.3 || ^8.0",

konradoboza: I am aware but constructor has strict typing. Shouldn't we also adjust property type as suggested then?

konradoboza: After internal sync and checking it now, it needs to stay as-is.

konradoboza (approved):

mateuszdebinski:
I created a new PR for 4.6 (https://github.com/ibexa/graphql/pull/73) as 3.3 is not longer supported

